PR #2, opened 2019-04-03
List modules freeipa-healthcheck depends on in setup.py

fcami:
f28 and up can use the automatic Python dependency generator if
required modules are listed in setup.py.

Source:
https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_dependencies
Example:
https://python-packaging.readthedocs.io/en/latest/dependencies.html

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,6 +12,12 @@ setup(
         'ipahealthcheck.ipa',
         'ipahealthcheck.meta'
     ],
+    install_requires=[
+          'ipaplatform',
+          'ipalib',
+          'ipaserver',
+          'ipapython',
+      ],
     entry_points={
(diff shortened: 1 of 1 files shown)


rcritten:
But will this pull in the insufficient PyPi version or use the system version? The PyPi version does not contain all imports that a full package installation does.

fcami:
Using "dnf localinstall" to install the resulting package pulls in the system packages providing those modules.

tiran:
* ``ipaserver`` is not packages on PyPI
* all other packages do not have a ``install`` subpackage

fcami:
Closing as invalid. Will manage dependencies directly in specfile.

PR #3, opened 2019-04-03
Error out of a source or check is requested and is not found

rcritten:
Otherwise it returns an empty list which makes it unclear
whether something was executed incorrectly or not at all.


fcami:
@rcritten +1, LGTM.

fcami:
Verified OK for both unknown source, and known source + unknown check.
Please feel free to merge.

PR #5, opened 2019-04-03
Use ipatopologysuffix-verify to validate replication topology

rcritten:
Both the ca and domain suffixes are evaluated and reported.

The results can be differentiated by the suffix of domain and ca
and the type of results, connection errors or too many agreements,
can be distinguished by 'type'.


fcami:
IPATopologyDomainCheck LGTM. Not tested on a real cluster.

PR #6, opened 2019-04-04
The RA agent certificate is tracked on all masters, not just CAs

rcritten:
The RA agent cert was only being checked in IPACertTracking when
a CA was configured causing it to display a false-positive of
an unknown certificate on masters without a CA.

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -55,18 +55,15 @@ def get_expected_requests(ca, ds, serverid):
     """
     template = paths.CERTMONGER_COMMAND_TEMPLATE
 
-    if ca.is_configured():
-        requests = [
-            {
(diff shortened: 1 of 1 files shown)


fcami:
@rcritten LGTM. Will confirm after running it locally.

fcami:
@rcritten confirmed. When replica has no CA I get the erroneous following output with HEAD:

# ipa-healthcheck --source ipahealthcheck.ipa.certs --check IPACertTracking --indent 1
 {
  "check": "IPACertTracking",
  "severity": 4,
  "uuid": "6c8a1d84-21df-4cdc-bf32-f98b989172d3",
  "when": "20190405201927Z",
  "duration": "0.323527",
  "kw": {
   "key": "20190405193131",
   "msg": "Unknown certmonger id 20190405193131"
  }
 },

The output is fixed with your patch. ACK :P

PR #9, opened 2019-04-09
Fix execution in CAless environment

rcritten:
The RA agent cert will be tracked on a master even if it
doesn't have an installed CA but it won't be there if there
is no CA at all in the topology.

Similarly, only check the ca toplogy suffix when a CA is
installed anywhere.


fcami:
LGTM. Not run, I don't have a complete CA-less cluster yet.

PR #10, opened 2019-04-09
Add a command-line option to print the available sources and checks

rcritten:
This will potentially make the --source and --check options easier
to discover.


fcami:
LGTM. Run locally, ACK.

PR #11, opened 2019-04-09
Use ipaplatform to determine service status

rcritten:
This makes this check more platform agnostic and re-uses the
same code IPA uses.

The service.is_installed() checks only to see if the service
is registered with systemd (on those platforms) so is insufficient
to tell whether a service is configured or not.

This also corrects an error where a failing service check was
not returning a result.


fcami:
LGTM. Run locally, confirmed working. ACK.

PR #12, opened 2019-04-10
Extend list of files to check including custodia, keytabs & tomcat

rcritten:
There is already a test for testing a basic FileCheck.

### DIFF
--- a/src/ipahealthcheck/ipa/files.py
+++ b/src/ipahealthcheck/ipa/files.py
@@ -2,16 +2,24 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 
+import logging
 import os
 
 from ipahealthcheck.core.files import FileCheck
 from ipahealthcheck.ipa.plugin import IPAPlugin, registry
 
+from ipalib import api, errors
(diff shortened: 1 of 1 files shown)


fcami:
When run on a CA-less replica, the Tomcat file check errors out and should not:
```
 {
  "source": "ipahealthcheck.ipa.files",
  "check": "TomcatFileCheck",
  "severity": 1,
  "uuid": "b885c5f0-38cf-4e12-8ba3-5fe879b71fab",
  "when": "20190411152142Z",
  "duration": "0.000070",
  "kw": {
   "exception": "'NoneType' object is not iterable"
  }
```

That's because of:
```
    def check(self):
        if not self.ca.is_configured():
            logger.debug('CA is not configured, skipping')
            return
```
which could be replaced by:
```
        if not self.ca.is_configured():
            logger.debug('CA is not configured, skipping')
            self.files = []
            return FileCheck.check(self)
```

rcritten:
I did the equivalent of that.

I could have also just done return () but this seems clearer.

fcami:
LGTM, ACK

PR #13, opened 2019-04-10
Ensure that no_schema is set when creating fake LDAP connections

rcritten:
This is necessary for testing when no IPA master is installed.

The "fake" connection is never actually used, the operations are
mocked away, but a real LDAPClient object is needed because of
assertions within the ipaldap code.


fcami:
LGTM

PR #14, opened 2019-04-11
Fix conversion of JSON to Result so that kw works in human output

rcritten:
When loading a JSON file it is converted into Result objects.
This was being done incorrectly for the kw data type. The entire
line was being passed in as kw instead of the kw directly.


fcami:
LGTM.
Verified by generating an output first, then calling
```
ipa-healthcheck --output-type human --input-file output
```
```
SUCCESS: ipahealthcheck.ipa.files.IPAFileCheck
```
becomes
```
SUCCESS: ipahealthcheck.ipa.files.IPAFileCheck._etc_named.keytab_owner
```

PR #15, opened 2019-04-14
Add a system plugin and a file system free space check.

fcami:
The system plugin is meant for operating-system level checks, like
checking for free space in file systems, amount of RAM and swap, etc.

The file system free space check is meant to avoid unfortunate ENOSPC
situations for DS and IPA.

### DIFF
--- a/src/ipahealthcheck/system/filesystemspace.py
+++ b/src/ipahealthcheck/system/filesystemspace.py
@@ -1,0 +1,81 @@
+#
+# Copyright (C) 2019 FreeIPA Contributors see COPYING for license
+#
+
+import shutil
+
+from ipahealthcheck.system.plugin import SystemPlugin, registry
+from ipahealthcheck.core.plugin import Result
+from ipahealthcheck.core import constants
+
+
+@registry
+class FileSystemSpaceCheck(SystemPlugin):
+    """
+    """
+
+    # watch important directories for FreeIPA
+    _pathchecks = {
+        '/var/lib/dirsrv/': 1024,
+        '/var/log/': 1024,
+        '/var/log/audit/': 512,
+        '/var/tmp/': 512,
+        '/tmp': 512
+    }
+
+    # File systems reaching 90% capacity risk fragmentation.
+    # Defragmentation is never desirable and not available
+    # on ext4 anyway. So error out at 20% free space.
+    min_free_percent = 20
+
+    def get_fs_free_space(self, pathname):
+        stat = shutil.disk_usage(pathname)
+        return int(stat.free / 2**20)
+
+    def get_fs_free_space_percentage(self, pathname):
+        stat = shutil.disk_usage(pathname)
+        return int(stat.free * 100 / stat.total)
+
+    def check(self):
+        for store in self._pathchecks:
+            percent_free = self.get_fs_free_space_percentage(store)
+            if percent_free < self.min_free_percent:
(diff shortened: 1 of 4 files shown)


fcami:
The check now generates an entry whether there is an error to report or not.

src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def check(self):
+        for store in self._pathchecks:
+            percent = self.get_fs_free_space_percentage(store)
+            if percent < self.min_free_percent:
+                yield Result(
+                    self, constants.ERROR,
+                    msg='%s: %s %s%% < %s%%' % (
+                        store, 'free space percentage under threshold:',
+                        percent, self.min_free_percent
```
rcritten: Add in the variables used to generate the data into msg as keywords, so store=store, percent=percent, min_free=self.min_free_percent.

src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -0,0 +1,79 @@
...
+                    )
+                )
+            free_space = self.get_fs_free_space(store)
+            threshold = self._pathchecks[store]
+            if free_space < threshold:
+                yield Result(
+                     self, constants.ERROR,
+                     msg='%s: %s %s MiB < %s MiB' % (
+                         store, 'free space under threshold:',
+                         free_space, threshold
```
rcritten: Add in the variables used to generate the data into msg as keywords, so store=store, free_space=free_space, threshold=threshold

src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -0,0 +1,79 @@
...
+                     )
+                )
+            else:
+                yield Result(
+                     self, constants.SUCCESS,
+                     msg='%s: %s %s MiB >= %s MiB' % (
```
rcritten: Other success don't return anything but passing but it doesn't hurt to pass additional data. You should include the variables used to calculate the outcome.

tests/test_system_filesystemspace.py
```diff
@@ -0,0 +1,39 @@
...
+from base import BaseTest
+from unittest.mock import Mock
+from util import capture_results
+from collections import namedtuple
+
+from ipahealthcheck.core import config, constants
+from ipahealthcheck.system.plugin import registry
+from ipahealthcheck.system.filesystemspace import FileSystemSpaceCheck
+
+
```
rcritten: Can you add a success test case as well?

src/ipahealthcheck/system/filesystemspace.py
fcami: Solved, I think.

src/ipahealthcheck/system/filesystemspace.py
fcami: Solved, I think.

tests/test_system_filesystemspace.py
fcami: Yes, doin' so

rcritten:
logging is imported but not used.

fcami:
re logging not used => removed.

tests/test_system_filesystemspace.py
fcami: Done. There are three tests now.

PR #16, opened 2019-04-17
Implement AD Trust checking

rcritten:
This checks a bunch of ADTrust options as outlined by Alexander.

The DS change is there because I set the wrong container originally and changing it to be right for AD Trust requires it be changed here as well. It's pedantic but could be important at some point.

### DIFF
--- a/src/ipahealthcheck/ipa/trust.py
+++ b/src/ipahealthcheck/ipa/trust.py
@@ -1,0 +1,499 @@
+#
+# Copyright (C) 2019 FreeIPA Contributors see COPYING for license
+#
+
+import logging
+import os
+import SSSDConfig
+
+from ipahealthcheck.ipa.plugin import IPAPlugin, registry
+from ipahealthcheck.core.plugin import Result
+from ipahealthcheck.core.plugin import duration
+from ipahealthcheck.core import constants
+
+from ipalib import api
+from ipaplatform.paths import paths
+from ipapython import ipautil
+from ipapython.dn import DN
+
+try:
+    from ipaserver.masters import ENABLED_SERVICE
+except ImportError:
+    from ipaserver.install.service import ENABLED_SERVICE
+try:
+    from ipapython.ipaldap import realm_to_serverid
+except ImportError:
+    from ipaserver.install.installutils import realm_to_serverid
+from ipaserver.install.adtrust import retrieve_netbios_name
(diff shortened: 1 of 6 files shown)


rcritten:
@fcami note that I'm toying with a new system for the response in the output to use a formatted string. If this looks ok we can apply it to the others as well (this will be important for localization eventually). It also forced me to be sure that all variables were listed in the keywords.

src/ipahealthcheck/ipa/trust.py
```diff
@@ -0,0 +1,499 @@
...
+                             msg='{key} service is not enabled')
+
+
+@registry
+class IPATrustControllerConfCheck(IPAPlugin):
+    """
+    Verify that smb.conf matches the template
+    """
+    @duration
+    def check(self):
```
abbra: It is just fine to have more options in the `net conf list` than what we have in the `smb.conf.template`.  Also, some options aren't really important to see exact. For example, `log level` might be any. I'm not sure how to achieve this...

src/ipahealthcheck/ipa/trust.py
```diff
@@ -0,0 +1,499 @@
...
+                             capture_output=True)
+        if result.returncode != 0:
+            yield Result(self, constants.ERROR,
+                         key='domain_list_error',
+                         sslctl=paths.SSSCTL,
+                         error=result.error_log,
+                         msg='Execution of {sslctl} failed: {error}')
+            return
+        sssd_domains = result.output.strip().split('\n')
+        if 'implicit_files' not in sssd_domains:
```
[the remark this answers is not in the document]
rcritten: Ok I can drop that check, though it would always pass :-)

src/ipahealthcheck/ipa/trust.py
rcritten: Should we punt on this then? Perhaps we can come up with a list of absolutely-required settings that I can validate.

rcritten:
The suggested changes applied plus a new mechanism to identify whether the current host is a trust agent or controller using servroles directly.

abbra:
LGTM. ACK.

src/ipahealthcheck/ipa/trust.py
abbra: well, there are few key ones. We can do them in a separate pull request. I'll build a list.

fcami:
LGTM too. ACK.

PR #17, opened 2019-04-17
Explicitly return SUCCESS instead of relying on core to do it

rcritten:
This lets us be sure that a result is going to be returned in the
tests.


fcami:
LGTM, runs OK, ACK

PR #18, opened 2019-04-17
Add systemd timer to run healthcheck daily and write log file

rcritten:
This will create a log file in /var/log/ipa/healthcheck at midnight.

The reason for the shell script is to have a configurable date.
systemd will not decode shell script variables and using bash -c
in the ExecStart line ipa-healthcheck was unable to communicate to
certmonger over DBus.

### DIFF
--- a/systemd/README
+++ b/systemd/README
@@ -1,0 +1,27 @@
+To enable daily execution using a systemd timer:
+
+Create the log directory
+
+# mkdir /var/log/ipa/healthcheck
+
+Copy the systemd configuration into place
+
+# cp ipa-healthcheck.timer /usr/lib/systemd/system
+# cp ipa-healthcheck.service /usr/lib/systemd/system
+
+Tell systemd about it:
+
+#systemctl daemon-reload
+
+Enable it:
+
+# systemctl enable ipa-healthcheck.timer
+# systemctl start ipa-healthcheck.timer
+
+Put a shell script in place to do the invocation:
+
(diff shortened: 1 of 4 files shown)


systemd/README
```diff
@@ -0,0 +1,23 @@
...
+
+Enable it:
+
+# systemctl enable ipa-healthcheck.timer
+# systemctl start ipa-healthcheck.timer
+
+Put a shell script in place to do the invocation:
+
+# cp ipa-healthcheck.sh /usr/libexec/ipa
+
```
fcami: missing instruction: "mkdir /var/log/ipa/healthcheck"

systemd/ipa-healthcheck.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+LOGDIR=/var/log/ipa/healthcheck/
+DATE=$(date +%Y%m%d)
+
+/usr/bin/ipa-healthcheck --output-file $LOGDIR/healthcheck.log-$DATE
```
fcami: nitpick: LOGDIR should be /var/log/ipa/healthcheck (no trailing /)
and line 5 should end with ${LOGDIR}/healthcheck.log-$DATE

Otherwise the command is constructed as "usr/bin/ipa-healthcheck --output-file /var/log/ipa/healthcheck//healthcheck.log-20190425" (double /)

rcritten:
Thanks, changes applied

PR #19, opened 2019-04-17
Make 'python3 setup.py test' work

rcritten:
Apply the steps for enabling testing using pytest provided at
https://docs.pytest.org/en/latest/goodpractices.html

This will make the check section in upstream packaging execute
the unit tests.


fcami:
LGTM, runs OK. ACK.

PR #20, opened 2019-04-25
Check more IPA-related services

fcami:
Check the status of ipa-custodia, ipa-dnskeysyncd,
ipa-otpd, kadmin and sssd.

This depends on these services being known to ipaplatform: https://github.com/freeipa/freeipa/pull/3075

### DIFF
--- a/src/ipahealthcheck/meta/services.py
+++ b/src/ipahealthcheck/meta/services.py
@@ -62,6 +62,43 @@ class httpd(ServiceCheck):
         return super(httpd, self).check()
 
 
+# services named with a hyphen cannot be addresses as knownservices.name
+# so use knownservices['name'] instead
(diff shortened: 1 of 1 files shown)


rcritten:
LGTM. Let's let the patch land upstream. May need to work out how we check for these or go ahead and create a 4-7 branch now to line up with what is in current releases freeIPA.

fcami:
Added exception handling for when services are not yet known to ipaplatform.
Added logging.
Output looks fine against freeipa-server-4.7.2-1.1.fc29.x86_64 which is missing (obviously) the changes from https://github.com/freeipa/freeipa/pull/3075

fcami:
https://github.com/freeipa/freeipa/pull/3075 hit master

PR #24, opened 2019-05-30
Report status if master is the CRL generator or renewal master

rcritten:
This will need to be aggregated elsewhere. It always reports as
success unless collecting the data fails.


rcritten:
Travis failed because the import is only in master. I'll add a skip if the import doesn't exist.

fcami:
Code LGTM and works well. Approved.

PR #25, opened 2019-06-03
Display text based severity instead of numeric

rcritten:
My intention originally was to save a few bits but it makes the
output much harder to read than I expected. There is no real
compelling reason to force users to translate on the analysis
end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,14 +96,14 @@ A basic check module consists of:
 
 A check yields a Result. This contains the outcome of the check including:
 
-* severity as defined in ipahealthcheck/core/constants.py
+* result as defined in ipahealthcheck/core/constants.py
 * msg containing a message to be displayed to the user.
 * kw, a python dictionary of name value pairs that provide details on the error
 
 The kw dict is meant to provide context for the check. Err on the side of
 too much information.
 
-msg and kw are optional if severity is SUCCESS.
+msg and kw are optional if result is SUCCESS.
 
 If a check consist of only a single test then it is not required to yield
 a Result, one marking the check as successful will be added automatically.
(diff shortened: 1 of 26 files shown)


fcami:
ERROR/SUCCESS does not ring like a severity to me. Severity would be high to low or 1-4 for instance. Error and Success sound more like a result but I'm not a native speaker.
So instead of:
```
    "severity": "SUCCESS",
 OR
    "severity": "ERROR",
```
I would like to see:
```
    "result": "SUCCESS",
 OR
    "result": "ERROR",
```

rcritten:
I think I'd like to reserve result in case we ever allow automatic repairs.
How about type, or response?

fcami:
'outcome' sounds appropriate if 'result' is reserved for automatic repairs.
Otherwise 'response' sounds fine to me.

rcritten:
On second thought result is the best. It is the result of the check, that's it. Making change now.

fcami:
Will review shortly, thanks

fcami:
PR LGTM, output looks good:
```
  {
    "source": "ipahealthcheck.meta.core",
    "check": "MetaCheck",
    "result": "SUCCESS",
    "uuid": "f2990145-3d03-476e-9553-40abc657fc0b",
    "when": "20190716203240Z",
    "duration": "0.000121",
    "kw": {
      "fqdn": "f29-idm0.laptop.example.org",
      "ipa_version": "4.7.91.dev201906201640+git970af1346",
      "ipa_api_version": "2.232"
    }
  },
  {
    "source": "ipahealthcheck.system.filesystemspace",
    "check": "FileSystemSpaceCheck",
    "result": "ERROR",
    "uuid": "41fd0451-4074-40ea-8310-9cf5fb8b2941",
    "when": "20190716203240Z",
    "duration": null,
    "kw": {
      "msg": "/var/lib/dirsrv/: free space percentage under threshold: 3% < 20%",
      "store": "/var/lib/dirsrv/",
      "percent_free": 3,
      "threshold": 20
    }
  },
```
I could not find any 'severity' left in the repository.
Approved.

Issue #26, opened 2019-06-12
If AD trust is enabled warn if package is not installed


rcritten:
If AD trust is enabled and the master does not have the freeipa-server-trust-ad package installed then the master will able to resolve users/groups via extdom plugin and sssd but won't be able to do framework-specific operations.

This should be a WARNING level issue.

PR #27, opened 2019-06-13
For trusts check to look ensure [free]ipa-server-trust-ad is installed

rcritten:
If the AD trust package isn't installed then the admin interfaces
won't be available. Lookups, etc will stil work so this could (and has)
lead to confusion

https://github.com/freeipa/freeipa-healthcheck/issues/26


abbra:
LGTM. Travis CI failed due to network issues accessing Fedora repos. I restarted it.

PR #28, opened 2019-06-14
Check if IPA is configured before launching tests.

fcami:


fcami:
When IPA is not configured:

```
# ipa-healthcheck --list-sources
Unable to initialize ipahealthcheck.dogtag: IPA is not configured on this system.
Unable to initialize ipahealthcheck.ds: IPA is not configured on this system.
Unable to initialize ipahealthcheck.ipa: IPA is not configured on this system.
ipahealthcheck.meta.services
   certmonger
   dirsrv
   gssproxy
   httpd
   ipa_custodia
   ipa_dnskeysyncd
   ipa_otpd
   kadmin
   krb5kdc
   named
   pki_tomcatd
   sssd
ipahealthcheck.meta.core
   MetaCheck
ipahealthcheck.system.filesystemspace
   FileSystemSpaceCheck
```

rcritten:
I wonder if we instead need to keep track of whether a plugin and/or registry is initialized and make decisions based on that.
So for example if the registry fails to initialize (e.g. because IPA isn't installed) then all plugins could be disabled. They would still list as sources but if you tried to call one it would return a message about being disabled, perhaps even why.
I'm less sure how this would all fit together, particularly at what level things would be stored. Could there be an ipa plugin, for example, that could execute even if the server is uninstalled? I don't know, not currently, but maybe in the future? That would toss in a wrench.

fcami:
After yesterday's discussion this is switched to having a proper error message if ipaserver is not configured.
It now includes a safety check for when ipa-healthcheck is installed and ipa-server packages are not. It might happen because ipa-healthcheck was installed from source or from packages outside of FreeIPA project control.

fcami:
Side note, I'm having doubts about what to send to stdout in this case. Sending "[{}]" seems more correct to me. @rcritten what do you think?

src/ipahealthcheck/core/main.py
```diff
@@ -13,14 +13,18 @@
...
-from ipahealthcheck.meta.services import ServiceCheck
-
-from ipaserver.install.installutils import is_ipa_configured
-
 
 logging.basicConfig(format='%(message)s')
 logger = logging.getLogger()
 
+try:
+    from ipahealthcheck.meta.services import ServiceCheck
```
rcritten: I wonder if it would read better to have the try/except just around is_ipa_configured and move the ServiceCheck import after this block. The ServiceCheck import will fail because it has imporst of its own so the is_ipa_configured will probably never get hit but it provides the clearest view of what is being attempted IMHO.

rcritten:
> Side note, I'm having doubts about:
> 
>     * what to send to stdout: sending "[{}]" or "[]" to stdout seems more correct to me than nothing as the json module cannot parse an empty string
> 
>     * which error code to use on exit. Even '0' sounds OK if we consider an empty healthcheck run OK 

I'm fine with no output and error 1 in this case. The tests have no real opportunity to execute.

fcami:
> > Side note, I'm having doubts about:
> > ```
> > * what to send to stdout: sending "[{}]" or "[]" to stdout seems more correct to me than nothing as the json module cannot parse an empty string
> > 
> > * which error code to use on exit. Even '0' sounds OK if we consider an empty healthcheck run OK 
> > ```
> 
> I'm fine with no output and error 1 in this case. The tests have no real opportunity to execute.

Thing is, if an external analysis tool expects proper json, 'nothing' is not. But that's a corner case.
OK @ error 1.

src/ipahealthcheck/core/main.py
fcami: So I remember why I didn't do it at first: I provides a clearer view, but then it triggers flake8:
```
src/ipahealthcheck/core/main.py:26:1: E402 module level import not at top of file
```
I've added a noqa flag for this line.

fcami:
- ipa-healthcheck now displays an error message if python3-ipaserver is not available and exits (code=1)
- if ipa is not installed, same behavior (with slightly different error message)

src/ipahealthcheck/core/main.py
```diff
@@ -13,12 +13,18 @@
...
 logging.basicConfig(format='%(message)s')
 logger = logging.getLogger()
 
+try:
+    from ipaserver.install.installutils import is_ipa_configured
+except ImportError:
+    logging.error("IPA server packages are not installed on this system.")
+    sys.exit(1)
+
+from ipahealthcheck.meta.services import ServiceCheck  # noqa: E402
```
rcritten: You can replace this with:

else:
    from ipahealthcheck.meta.services import ServiceCheck

rcritten:
ack.

PR #33, opened 2019-06-24
Return non-zero on error

rcritten:
Fix one check that was failing due to API issue when run using IPA 4.7.2. This failure had the side effect of showing that on failures a non-zero return value wasn't being set. Fix that as well.


fcami:
LGTM. Ran without issues. ACK.

fcami:
Closing as https://github.com/freeipa/freeipa-healthcheck/pull/33 is merged.

Issue #29, opened 2019-06-20
Parse /etc/ipa/ca.crt and look for expired or expiring CA certificates


rcritten:
Use ipalib.x509::load_certificate_list_from_file() to load /etc/ipa/ca.crt and look at the cert expiration dates and warn.

fcami:
Fixed by https://github.com/freeipa/freeipa-healthcheck/pull/70 - Closing.

PR #34, opened 2019-06-28
ipa: look for expired or expiring CA certificate

fcami:
WIP, this is missing a test.

https://github.com/freeipa/freeipa-healthcheck/issues/29

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -879,3 +879,35 @@ class IPACertmongerCA(IPAPlugin):
             else:
                 yield Result(self, constants.SUCCESS,
                              key=ca)
+
+
+@registry
+class IPAMainCAExpirationCheck(IPAPlugin):
+
+    @duration
+    def check(self):
+        try:
+            cert = x509.load_certificate_from_file(paths.IPA_CA_CRT)
+        except IOError as e:
+            logger.debug("Could not open %s!" % paths.IPA_CA_CRT)
+            yield Result(self, constants.ERROR,
+                         key=paths.IPA_CA_CRT,
+                         msg='Error opening IPA main CA at %s: %s' %
+                         (paths.IPA_CA_CRT, e))
+            return
+        except (TypeError, ValueError):
+            logger.debug("% is not a valid certificate!" % paths.IPA_CA_CRT)
+            yield Result(self, constants.ERROR,
+                         key=paths.IPA_CA_CRT,
+                         msg='IPA main CA %s is invalid.' % paths.IPA_CA_CRT)
+            return
+        now = datetime.datetime.now() + datetime.timedelta(weeks=2)
+        if cert.not_valid_after <= now:
+            logger.debug("% is expiring soon!" % paths.IPA_CA_CRT)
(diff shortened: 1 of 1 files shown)


src/ipahealthcheck/ipa/certs.py
```diff
@@ -879,3 +879,35 @@ def check(self):
...
                              key=ca)
+
+
+@registry
+class IPAMainCAExpirationCheck(IPAPlugin):
+
+    @duration
+    def check(self):
+        try:
+            cert = x509.load_certificate_from_file(paths.IPA_CA_CRT)
```
rcritten: This should be load_certificate_list_from_file() otherwise only the first cert will be loaded. We want to check the entire chain.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -879,3 +879,35 @@ def check(self):
             else:
                 yield Result(self, constants.SUCCESS,
                              key=ca)
+
+
+@registry
+class IPAMainCAExpirationCheck(IPAPlugin):
```
rcritten: I see where you're going with the Main but I think I'd drop it and maybe add Chain: IPACAChainExpirationCheck

src/ipahealthcheck/ipa/certs.py
```diff
@@ -879,3 +879,35 @@ def check(self):
...
+                         msg='Error opening IPA main CA at %s: %s' %
+                         (paths.IPA_CA_CRT, e))
+            return
+        except (TypeError, ValueError):
+            logger.debug("% is not a valid certificate!" % paths.IPA_CA_CRT)
+            yield Result(self, constants.ERROR,
+                         key=paths.IPA_CA_CRT,
+                         msg='IPA main CA %s is invalid.' % paths.IPA_CA_CRT)
+            return
+        now = datetime.datetime.now() + datetime.timedelta(weeks=2)
```
rcritten: I realize it's WIP but the delta should be based on self.config.cert_expiration_days. In fact, it might not be unreasonable to have the CA checking much longer, 30+ days, since external renewal will rely on someone else doing work which can take a while.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -879,3 +879,35 @@ def check(self):
...
+        if cert.not_valid_after <= now:
```
rcritten: I also wonder if we should abstract expiration so the same function can be used in both the system cert expiration and this.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -879,3 +879,35 @@ def check(self):
...
+            logger.debug("% is expiring soon!" % paths.IPA_CA_CRT)
+            yield Result(self, constants.ERROR,
+                         key=paths.IPA_CA_CRT,
+                         msg='IPA main CA \'%s\' is expiring soon.' %
+                         paths.IPA_CA_CRT)
+        else:
+            yield Result(self, constants.SUCCESS,
+                         key=paths.IPA_CA_CRT)
```
rcritten: Yeah, this probably works. I like showing our work :-)

Issue #37, opened 2019-07-03
Human output can return a KeyError for certmonger tracking issues


rcritten:
If an expected tracking cert is not found in certmonger then the request is added as a string to the msg kw.

In human output this is displayed as msg.format(**kw) which will raise KeyError: "'cert-database'" because the literal string in msg the converted dict and {} are being interpreted as a variable.

rcritten:
https://github.com/freeipa/freeipa-healthcheck/pull/39

PR #39, opened 2019-07-03
When reporting certmonger request issues flatten the request

rcritten:
This was previously using str() to convert the dict into a string.
In Output this could be passed into string.format() which will
interpret the embedded {} of the dict as variable names.

Instead flatten the dictionary into a comma-separate string of
name/value pairs

https://github.com/freeipa/freeipa-healthcheck/issues/37

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ install:
 
 script:
     - tox -epep8,flake8
-    - docker run -v ${TRAVIS_BUILD_DIR}:/root/src/ fedora:29 /bin/sh -c "dnf -y install freeipa-server tox; cd /root/src; tox -epy3"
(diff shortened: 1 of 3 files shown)


fcami:
LGTM. Sample output:
```
  {
    "check": "IPACertTracking",
    "severity": 2,
    "uuid": "2c2e044f-8c4c-4e78-b186-9fd1e341c2ce",
    "when": "20190716172146Z",
    "duration": "0.590354",
    "kw": {
      "msg": "Missing tracking for cert-file=/var/kerberos/krb5kdc/kdc.crt, key-file=/var/kerberos/krb5kdc/kdc.key, ca-name=None, cert-postsave-command=/usr/libexec/ipa/certmonger/renew_kdc_cert"
    }
  },
```

fcami:
Also thanks for the travis fix!

PR #41, opened 2019-07-17
Add service dependencies to checks that rely on services to be up

rcritten:
Some checks require one or more services to be running in order to execute. Add dependencies on those services so those checks are skipped if the dependency is not running.
For example, if dirsrv is not running there is no point in returning a lot of failures that LDAP can't be contacted.

This also fixes an issue from a previous commit switching to a string result value.


fcami:
Verified:
- shutting down dirsrv and/or krb5kdc disables some tests
- with --debug, when a dependency is not present an error message is shown
ACK

Issue #35, opened 2019-07-02
add log rotation


fcami:
As it stands healthcheck logs have no rotation.
I suppose adding a file in /etc/logrotate.d/ is the right way to do it.

PR #42, opened 2019-07-18
Switch logs to logrotate (#35)

fcami:
According to anacron logrotate starts at 00:05 at the earliest and  00:50 at the latest.
Make sure healthcheck.timer invokes healthcheck long after logrotate is done.
+
Switch to logging into /var/log/ipa/healthcheck/healthcheck.log so that the latest log is always consistently named.
+ 
Add logrotate default config (#35).


fcami:
Hi @rcritten this seems to pass our testing. The only question that remains is, is 14 day retention enough or do we want more (28?)

fcami:
As per discussion updated to 30 days.

README.md
```diff
@@ -16,16 +16,50 @@ The major areas currently covered are:
...
+The simplest way to use Healthcheck is to run it from the command-line as root as ipa-healthcheck. Running from the command-line will display the output to the console unless --output-file=FILENAME is used.
+There is output for _all_ tests so we can be sure that an error condition isn't providing a false positive. The command-line option --failures-only will skip printing the SUCCESS conditions.
 
-It can be run from the command-line as root as ipa-healthcheck. Running from the command-line by default will display the output to the console.
+To automate running Healthcheck every day a systemd timer can be used. 
+The default destination directory for healthcheck logs is `/var/log/ipa/healthcheck` and this can be the input into a monitoring system to track changes over time or to alert if a test goes from working to error or warning.
 
-There is output for _all_ tests so we can be sure that an error condition isn't providing a false positive. The command-line option --failures-only will skip printing the SUCCESS conditions.
+A systemd timer is shipped by RHEL and Fedora packages but not enabled. To enable it:
```
rcritten: The timer can be shipped on other distros as well. I'd change it to something like "A systemd timer is provided but is not enabled by default. To enable it:"

README.md
```diff
@@ -16,16 +16,50 @@ The major areas currently covered are:
...
+
+Tell systemd about it and enable it:
+
+    # systemctl daemon-reload
+    # systemctl enable ipa-healthcheck.timer
+    # systemctl start ipa-healthcheck.timer
+
+Finally add a proper logrotate configuration:
+
+    # cp logrotate/ipahealthcheck /etc/logrotate.d/
```
rcritten: It's worth mentioning that logrorate relies on crond so that needs to be installed and enabled as well.

README.md
fcami: Fixed, thanks!

README.md
fcami: Fixed.

Issue #44, opened 2019-07-18
when root has an expired ticket ipa-healthcheck service fails


fcami:
when root has an expired ticket:
```
# klist
Ticket cache: KCM:0
Default principal: admin@LAPTOP.EXAMPLE.ORG
 
Valid starting       Expires              Service principal
06/24/2019 16:21:07  06/25/2019 16:21:03  HTTP/f29-idm0.laptop.example.org@LAPTOP.EXAMPLE.ORG
06/24/2019 16:21:05  06/25/2019 16:21:03  krbtgt/LAPTOP.EXAMPLE.ORG@LAPTOP.EXAMPLE.ORG
```

starting the service with 'systemctl start ipa-healthcheck' leads to:
```
# systemctl status ipa-healthcheck
● ipa-healthcheck.service - Execute IPA Healthcheck
   Loaded: loaded (/usr/lib/systemd/system/ipa-healthcheck.service; disabled; vendor preset: disabled)
   Active: failed (Result: exit-code) since Thu 2019-07-18 23:43:46 CEST; 11min ago
  Process: 29703 ExecStart=/usr/libexec/ipa/ipa-healthcheck.sh (code=exited, status=1/FAILURE)
 Main PID: 29703 (code=exited, status=1/FAILURE)
 
Jul 18 23:43:38 f29-idm0.laptop.example.org systemd[1]: Started Execute IPA Healthcheck.
Jul 18 23:43:46 f29-idm0.laptop.example.org python3[29704]: GSSAPI client step 1
Jul 18 23:43:46 f29-idm0.laptop.example.org python3[29704]: GSSAPI client step 1
Jul 18 23:43:46 f29-idm0.laptop.example.org python3[29704]: GSSAPI Error: Unspecified GSS failure.  Minor code may provide more information (Ticket expired)
Jul 18 23:43:46 f29-idm0.laptop.example.org systemd[1]: ipa-healthcheck.service: Main process exited, code=exited, status=1/FAILURE
Jul 18 23:43:46 f29-idm0.laptop.example.org systemd[1]: ipa-healthcheck.service: Failed with result 'exit-code'.
```

fcami:
set KRB5_CLIENT_KTNAME in the shell script maybe? thanks @rcritten

fcami:
The following diff fixes it:
```
# diff -U0 main.py-orig main.py
--- main.py-orig	2019-07-18 23:03:00.000000000 +0200
[...]

PR #46, opened 2019-07-19
Add KRB5CCNAME management (#44)

fcami:
When root has an expired kerberos TGT the ipa-healthcheck service fails
with "GSSAPI Error: Unspecified GSS failure (...) (Ticket expired)".
Set a random KRB5CCNAME so that it does not happen and cleanup after
ourselves.


fcami:
Closing in favor of https://github.com/freeipa/freeipa-healthcheck/pull/47 which is much simpler.

PR #47, opened 2019-07-22
Set KRB5CCNAME and KRB5_CLIENT_KTNAME early (#44)

fcami:
When root has an expired kerberos TGT the ipa-healthcheck service fails
with "GSSAPI Error: Unspecified GSS failure (...) (Ticket expired)".
Set KRB5CCNAME to MEMORY just like kadmin does.
This requires in turn setting KRB5_CLIENT_KTNAME to initialize the
credential cache properly.


fcami:
Good catch, thanks @rcritten . PR updated.

PR #48, opened 2019-07-22
constants.py: reorder and renumber error constants

fcami:

### DIFF
--- a/src/ipahealthcheck/core/constants.py
+++ b/src/ipahealthcheck/core/constants.py
@@ -6,15 +6,15 @@ DEFAULT_OUTPUT = 'json'
 
 # Error reporting result
 SUCCESS = 0
-CRITICAL = 1
-ERROR = 2
-WARNING = 4
+WARNING = 10
+ERROR = 20
+CRITICAL = 30
 
 _levelToName = {
     SUCCESS: 'SUCCESS',
-    CRITICAL: 'CRITICAL',
-    ERROR: 'ERROR',
     WARNING: 'WARNING',
+    ERROR: 'ERROR',
+    CRITICAL: 'CRITICAL',
 }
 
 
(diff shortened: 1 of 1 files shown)


src/ipahealthcheck/core/constants.py
```diff
@@ -6,15 +6,15 @@
 
 # Error reporting result
 SUCCESS = 0
-CRITICAL = 1
-ERROR = 2
-WARNING = 4
+WARNING = 10
+ERROR = 20
+CRITICAL = 40
```
rcritten: This should be 30.

src/ipahealthcheck/core/constants.py
```diff
@@ -6,15 +6,15 @@
...
 
 _levelToName = {
     SUCCESS: 'SUCCESS',
-    CRITICAL: 'CRITICAL',
-    ERROR: 'ERROR',
     WARNING: 'WARNING',
+    ERROR: 'ERROR',
+    CRITICAL: 'CRITICAL',
```
rcritten: Ala logging we should probably add since we're in here:

'CRITICAL': CRITICAL,
'ERROR': ERROR

etc to translate string to number.

src/ipahealthcheck/core/constants.py
```diff
@@ -6,27 +6,43 @@
...
     """
     name = _levelToName.get(level)
     if name is not None:
         return name
-    else:
-        return level
+    name = _nameToLevel.get(level)
+    if name is not None:
+        return name
+    return level
```
rcritten: If you want to keep the numeric and string constants separate (which is fine) you can simplify the code to:

return _levelToName.get(level) or _nameToLevel.get(level) or level

fcami: Funny, I had looked at logging. Too much copy&paste.

PR #49, opened 2019-07-22
Use 2 as the default JSON indention

rcritten:
The output was just too huge for some editors and it was generally
useless to paw through a single-line 40k file manually.


fcami:
LGTM

Issue #50, opened 2019-07-22
input-file throws an exception


rcritten:
```
# ipa-healthcheck --input-file /var/log/ipa/healthcheck/healthcheck.log --output-type human 2>&1
[proper results]
Traceback (most recent call last):
  File "/usr/bin/ipa-healthcheck", line 11, in <module>
    load_entry_point('ipahealthcheck==0.3', 'console_scripts', 'ipa-healthcheck')()
  File "/usr/lib/python3.7/site-packages/ipahealthcheck/core/main.py", line 224, in main
    for result in results.results:
AttributeError: 'NoneType' object has no attribute 'results'

```
The reason is there are no real results to see what the exit code should be.
Reconsidering this it would be nice if input-file wasn't limited to the type of output. It would be handy to be able to find failures in older runs using this as well, in the JSON output type too.
Being able to also limit to source and check would be nice as well. The tricky bit is that for the normal case only those sources/checks are executed but in the loading from file the output needs to be parsed.

PR #52, opened 2019-07-22
Make input-file globally available

rcritten:
There is value in being able to re-process existing log files for
any output type so move input-file out of Human and process it
in main.

The main difference in processing is where the results come from:
either a file or by running the checks. Post-processing works
the same.

This will allow one to use an existing log to:
- look for only failures
- look at a specific source or check
- output into any format

https://github.com/freeipa/freeipa-healthcheck/issues/50


fcami:
LGTM

Issue #53, opened 2019-07-23
RFE: implement --output-file FILENAME for --output-type human


fcami:
--output-file FILENAME only works with --output-type json

It would be nice to have it working with both types of output.

PR #55, opened 2019-07-23
Make --output-file globally available

rcritten:
There is valid in being able to provide an output file regardless
of the output type so move output-file out of JSON and process it
in the Output class itself.

An Output plugin only needs to implement the generate() method
to provide a string to be written from the available results.

https://github.com/freeipa/freeipa-healthcheck/issues/53


fcami:
LGTM, works OK.

PR #56, opened 2019-07-26
Add an option to search by specific failure severity

rcritten:
A WARNING is not a failure. Drop it from --failures only.

Add an option to search by specific failure severity.

### DIFF
--- a/man/man1/ipa-healthcheck.1
+++ b/man/man1/ipa-healthcheck.1
@@ -40,7 +40,10 @@ Execute this particular check within a source. A source must be supplied as well
 Set the output type. The default is JSON.
 .TP
 \fB\-\-failures\-only\fR
-Only report failures. Drop the successful results.
+Exclude SUCCESS results on output.
+.TP
+\fB\-\-severity=\fRSEVERITY\fR
+Only report errors in the requested severity of SUCCESS, WARNING, ERROR or CRITICAL. This can be provided multiple times to search on multiple levels.
 .TP
 \fB\-\-debug\fR
 Generate additional debugging output.
--- a/src/ipahealthcheck/core/main.py
+++ b/src/ipahealthcheck/core/main.py
@@ -163,7 +163,10 @@ def parse_options(output_registry):
                         help='File to read as input')
     parser.add_argument('--failures-only', dest='failures_only',
                         action='store_true', default=False,
-                        help='Exclude SUCCESS result on output')
+                        help='Exclude SUCCESS results on output')
+    parser.add_argument('--severity', dest='severity', action="append",
+                        help='Include only the selected severity(s)',
+                        choices=[key for key in constants._nameToLevel])
     for plugin in output_registry.plugins:
         onelinedoc = plugin.__doc__.split('\n\n', 1)[0].strip()
         group = parser.add_argument_group(plugin.__name__.lower(),
--- a/src/ipahealthcheck/core/output.py
+++ b/src/ipahealthcheck/core/output.py
@@ -3,7 +3,7 @@
 #
 
 import json
-from ipahealthcheck.core.constants import getLevelName
+from ipahealthcheck.core.constants import getLevelName, _nameToLevel, SUCCESS
 from ipahealthcheck.core.plugin import Registry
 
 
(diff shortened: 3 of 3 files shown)


fcami:
LGTM. One comment is that I'd feel better if we had a test for each CLI knob ( https://github.com/freeipa/freeipa-healthcheck/issues/57 ).

fcami:
So after thinking about it, I'm not so sure filtering WARNINGs from --failures-only is fine.

Consider the following WARNING:
```
  {
    "source": "ipahealthcheck.ipa.files",
    "check": "TomcatFileCheck",
    "result": "WARNING",
    "uuid": "c05bdaea-a53d-48a6-a152-e88c97a5a7d4",
    "when": "20190729095557Z",
    "duration": "0.000245",
    "kw": {
      "key": "_etc_pki_pki-tomcat_server.xml_mode",
      "path": "/etc/pki/pki-tomcat/server.xml",
      "type": "mode",
      "expected": "0660",
      "got": "0000",
      "msg": "Permissions of /etc/pki/pki-tomcat/server.xml are 0000 and should be 0660"
    }
  },
```
Since it's a warning:
```
# ipa-healthcheck --failures-only
[]
```
But (obviously) this leads into pki-tomcatd not starting after "ipactl restart":
```
# systemctl status pki-tomcatd@pki-tomcat.service
● pki-tomcatd@pki-tomcat.service - PKI Tomcat Server pki-tomcat
   Loaded: loaded (/lib/systemd/system/pki-tomcatd@.service; enabled; vendor preset: disabled)
  Drop-In: /etc/systemd/system/pki-tomcatd@pki-tomcat.service.d
           └─ipa.conf
   Active: failed (Result: exit-code) since Mon 2019-07-29 11:58:01 CEST; 51s ago
  Process: 1790 ExecStop=/usr/libexec/tomcat/server stop (code=exited, status=1/FAILURE)
  Process: 1421 ExecStartPost=/usr/libexec/ipa/ipa-pki-wait-running (code=exited, status=0/SUCCESS)
  Process: 1420 ExecStart=/usr/libexec/tomcat/server start (code=exited, status=143)
  Process: 1261 ExecStartPre=/usr/bin/pkidaemon start pki-tomcat (code=exited, status=0/SUCCESS)
  Process: 2153 ExecStartPre=/usr/sbin/pki-server migrate --instance pki-tomcat (code=exited, status=1/FAILURE)
 Main PID: 1420 (code=exited, status=143)

Jul 29 11:58:00 f29-idm0.laptop.example.org systemd[1]: Starting PKI Tomcat Server pki-tomcat...
Jul 29 11:58:01 f29-idm0.laptop.example.org pki-server[2153]: ERROR: Error reading file '/etc/pki/pki-tomcat/server.xml': failed to load external entity "/etc/pki/pki-tomcat/serv>
Jul 29 11:58:01 f29-idm0.laptop.example.org systemd[1]: pki-tomcatd@pki-tomcat.service: Control process exited, code=exited status=1
Jul 29 11:58:01 f29-idm0.laptop.example.org systemd[1]: pki-tomcatd@pki-tomcat.service: Failed with result 'exit-code'.
Jul 29 11:58:01 f29-idm0.laptop.example.org systemd[1]: Failed to start PKI Tomcat Server pki-tomcat.
```
This is because core/files.py uses "if mode != fmode:" and should use a more precise mode comparison: if fmode is more permissive that fmode, yield WARNING, elif fmode is less permissive than fmode, yield ERROR. 

Obviously restarting ipa leads to a non-functional pki-tomcatd and then ipa-healthcheck catches some errors, but I would argue this is too late.

A related logic applies to IPACertmongerExpirationCheck where WARNING is used to catch expiring certificates. An admin using --failures-only would not catch the obvious action item unless it's too late, incurring downtime.

However the --severity filter is useful and nice.

Can you split this PR in two?

rcritten:
Converted this PR into only adding the ability to query based on severity level. We can tackle how to handle warnings in the future. My concern was with out to deal with false positives.

fcami:
hi @rcritten LGTM. I'm wondering whether we should change the logic - see comment above. But that's not mandatory.

rcritten:
Made --severity multi-valued so one can search on ERROR and CRITICAL, for example.

fcami:
Code LGTM, man page matches actual behavior. ACK.

Issue #60, opened 2019-07-30
Change DNA plugin to return WARNING if no range is set


rcritten:
Current ipahealthcheck.ipa.dna returns SUCCESS both when a range is set and when one is not set.

Elevate the not set version to WARNING so admins can be aware in advance that with no range then uid/gid cannot be allocated from this master. This is particularly important if the only master with a range defined is retired.

PR #61, opened 2019-07-30
Change DNA no range from SUCCESS to WARNING

rcritten:
Elevate the not set version to WARNING so admins can be aware in
advance that with no range then uid/gid cannot be allocated from
this master. This is particularly important if the only master with
a range defined is retired.

https://github.com/freeipa/freeipa-healthcheck/issues/60


fcami:
LGTM.

PR #67, opened 2019-08-08
Enhance man page with a pointer to README.md and websites

fcami:
Add a note to ipa-healthcheck(1) so that users find README.md which contains detailed information about executed checks.


rcritten:
Looks good, thanks.

PR #68, opened 2019-08-08
Lookup AD user by SID and not by hardcoded username

rcritten:
Looking up the user user as Administrator@REALM is not
portable because the administrator login name may be
localized.


fcami:
Code LGTM, both commit work well. Ack.

Issue #71, opened 2019-09-18
error when no tty


rcritten:
```
$ ssh  -i  config/id_rsa -o StrictHostKeyChecking=no
root@test.example.com ipa-healthcheck
Output raised OSError: [Errno 6] No such device or address: '/dev/tty'
```

It works correctly with an option `-t`:
```
$ ssh -t  -i  config/id_rsa -o StrictHostKeyChecking=no
root@test.example.com ipa-healthcheck
[
  {
    "source": "ipahealthcheck.meta.services",
    "check": "certmonger",
    "result": "SUCCESS",
    "uuid": "b9a530a8-82df-4690-b789-39d5a3cd3455",
    "when": "20190918140724Z",
    "duration": "0.016964",
    "kw": {
      "status": true
    }
  },
<snip>
```

rcritten:
Probably need to use isatty instead of explicit check for /dev/tty in output.py

PR #74, opened 2019-09-24
Use sys.stdout by default as /dev/tty does not always exist (#71)

fcami:
Healthcheck currently defaults to using /dev/tty.
/dev/tty does not exist when launching commands non-interactively,
for instance with ssh:
$ ssh root@ipa.test ipa-healthcheck
Output raised OSError: [Errno 6] No such device or address: '/dev/tty'

Switch to using sys.stdout in write_file() when --output-file is
not used.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/71


fcami:
I'll update the PR, thx.

Issue #76, opened 2019-09-30
Implement CA range checking


rcritten:
Blog post about Dogtag range management that might help understand how to check this:

https://frasertweedale.github.io/blog-redhat/posts/2019-07-26-dogtag-replica-ranges.html

rcritten:
The CA is moving away from sequential serial numbers to RSN. Dropping this.

PR #77, opened 2019-09-30
Implement simple CA range checking

rcritten:
Checking the CS.cfg ranges requires visibility for the whole
cluster so only report those values.

We can check that the data stored in LDAP doesn't overlap and
that the nextRange value is defined.

https://github.com/freeipa/freeipa-healthcheck/issues/76


rcritten:
@frasertweedale can you take a look?

src/ipahealthcheck/dogtag/range.py
```diff
@@ -0,0 +1,170 @@
...
+        except errors.NotFound:
+            entries = []
+
+        if entries:
+            if len(entries) != 1:
+                # Can't really proceed if there are multiple so fail loudly.
+                yield Result(self, constants.CRITICAL,
+                             key='nextRange',
+                             dn='ou=certificateRepository,ou=ca,o=ipaca',
+                             msg='multiple {key} values in {dn}')
```
frasertweedale: This seems to also handle the `len(entries) == 0` case, so perhaps the error message should mention the number of entries (and that only one entry was expected).

src/ipahealthcheck/dogtag/range.py
```diff
@@ -0,0 +1,170 @@
...
+                return
+            else:
+                nextrange = int(entries[0].single_value.get('nextRange', 0))
```
frasertweedale: Certificate repository (and KRA keyRepository) uses hex encoding.  Even though they may look decimal-ish :)

src/ipahealthcheck/dogtag/range.py
```diff
@@ -0,0 +1,170 @@
...
+
+        yield Result(self, constants.SUCCESS,
+                     key='nextRange',
+                     nextrange=nextrange)
+
+        filterstr = "(objectclass=pkiRange)"
+        attrlist = ['beginRange', 'endRange', 'host']
+        try:
+            entries = self.conn.get_entries(
+                DN('ou=requests,ou=ranges,o=ipaca'),
```
frasertweedale: So here we are comparing request ranges against the certificateRepository `nextRange`.  We need to output and compare range entries and `nextRange` values separate for each of:

- CA certificateRepository
- CA requests
- CA replicas
- KRA keyRepository
- KRA requests
- KRA replicas

It's a pain, I know.  I suppose the best way is a superclass with all the logic, and a subclass for each managed range that just plugs in a few parameters (what are the CS.cfg params, repository object DN, range object base DN, whether the numbers are decimal or hex).

src/ipahealthcheck/dogtag/range.py
```diff
@@ -0,0 +1,170 @@
...
+
+        if not self.ca.is_configured():
+            logger.debug("No CA configured, skipping dogtag config check")
+            return
+
+        # (key_name, required)
+        keys = [('dbs.beginSerialNumber', True),
+                ('dbs.endSerialNumber', True),
+                ('dbs.nextBeginSerialNumber', False),
+                ('dbs.nextEndSerialNumber', False)]
```
frasertweedale: And yeah, same comment here, this looks fine for serials, but we need to do similar thing for all the other managed ranges.

src/ipahealthcheck/dogtag/range.py
rcritten: The no entries case is handled on line 67. The message for multiple can be improved, I'll push an update.

src/ipahealthcheck/dogtag/range.py
rcritten: Ah ok, I didn't get that part from the docs.

rcritten:
Sure, you can take it over. I'll push the one minor update for the multiple keys message.

rcritten:
I pushed a fix for multiple ranges and added a test case for it.

Issue #30, opened 2019-06-20
Look for dangling RUVs


rcritten:
RUVs of old replicas can cause issues with replication. See if there is a way to identify them so they can be removed (or recommend running cleanallruv).

rcritten:
Finding dangling RUVs requires connecting to all masters to get their replicaid. healthcheck is local only so the best we can do is report the local RUVs and let an outside process analyse it.

PR #78, opened 2019-10-01
Report the CS and DS RUV values

rcritten:
It isn't possible to identify dangling RUVs since healthcheck
runs locally only. The best we can do is report the current
value(s).

https://github.com/freeipa/freeipa-healthcheck/issues/30


fcami:
Code LGTM and indeed I don't think we can do better in a single-host fashion. ACK

PR #79, opened 2019-10-01
Report correct error when running a source and a req. isn't satisfied

rcritten:
If a requirement isn't satisfied when running a plugin then it
is skipped and nothing is reported. This previously reported the
confusing error "Source <foo> not found".

Run through the list again in this case to determine if a
dependency is missing and report a more correct error.


rcritten:
To test run:
```
# ipactl stop
# ipa-healthcheck --source ipahealthcheck.ipa.dna
```
Without the patch it will fail with Source not found.

fcami:
Change LGTM, verified. ACK

PR #80, opened 2019-10-08
Validate expected DNS SRV records based on server roles

rcritten:
IPA can provide the expected list of SRV records based on its
configuration (using the IPASystemRecords class) via

ipa dns-update-system-records --dry-run

Take the equivalent of that output and compare it to what is in
DNS.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setup(
         'ipahealthcheck.ipa': [
             'ipacerts = ipahealthcheck.ipa.certs',
             'ipadna = ipahealthcheck.ipa.dna',
+            'ipadns = ipahealthcheck.ipa.dns',
             'ipafiles = ipahealthcheck.ipa.files',
             'ipahost = ipahealthcheck.ipa.host',
             'iparoles = ipahealthcheck.ipa.roles',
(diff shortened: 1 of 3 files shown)


ofalk:
Great addition - I was looking for this (https://pagure.io/freeipa/issue/7871).

I didn't fully analyse the code, therefore I have one question - excuse my laziness.
Does this code check all records on _ALL_ DNS servers?
This was my main problem in the past. The first DNS server correctly showed the records, but the second and third DNS server had missing records and therefore we had some 'funny' side effects.
Maybe we want an option to specify the DNS server to be checked?

rcritten:
I'm not sure how possible that is. healthcheck runs in the context of a single master and relies on the local resolver. I'm not sure how realistic it would be to provide a list of DNS masters to validate against or if that would scale.

fcami:
Tested OK. Code LGTM. Checking on ALL DNS servers should be done externally to healthcheck which is a single-system check. Hence ACK.

ofalk:
Sorry for getting back late.
Do you think it would be feasible to provide a list of DNS servers to check for the records?

PR #81, opened 2019-10-21
Minor Fixes

rcritten:
I backported healthcheck to IPA 4.6.4 in Fedora 27 and discovered a number of issues that affect the master branch as well. This tidies a few things up.


fcami:
LGTM, but the reason of the first change to  src/ipahealthcheck/ipa/certs.py is not immediately obvious to me. @rcritten could you explain please?

rcritten:
For the log message it in the HTTP case we load the cert then check to see if the issuer matches. in the DS case it checks directly in the db without loading a local copy of the cert.

The second change in the file is because cert is loaded but never used.

fcami:
Thanks for the explanation. ACK.

PR #82, opened 2019-11-06
When collecting filesystem info if a path doesn't exist then ignore

rcritten:
This was causing false positivies in containers where some file
systems, notably /var/log/audit, is not mounted.

If a filesystem isn't available then report as SUCCESS with a
useful message.

### DIFF
--- a/src/ipahealthcheck/system/filesystemspace.py
+++ b/src/ipahealthcheck/system/filesystemspace.py
@@ -2,11 +2,27 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 
+import os
 import shutil
 
 from ipahealthcheck.system.plugin import SystemPlugin, registry
 from ipahealthcheck.core.plugin import duration, Result
 from ipahealthcheck.core import constants
+
+
+def in_container():
+    """Determine if we're running in a container."""
+    with open('/proc/1/sched', 'r') as sched:
+        data = sched.readline()
+
+    checks = [
+        data.split()[0] not in ('systemd', 'init',),
+        os.path.exists('/.dockerenv'),
+        os.path.exists('/.dockerinit'),
+        os.getenv('container', None) is not None
+    ]
+
+    return any(checks)
 
 
 @registry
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -2,13 +2,29 @@
...
+def in_container():
+    """Determine if we're running in a container."""
+    with open('/proc/1/sched', 'r') as sched:
+        data = sched.readline()
+
+    checks = [
+        data.split()[0] not in ('systemd', 'init',),
+        os.path.exists('/.dockerenv'),
+        os.path.exists('/.dockerinit'),
+        os.getenv('container', None) is not None
```
fcami: That should work on k8s/ocp + podman

src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -2,13 +2,29 @@
...
 
+def in_container():
+    """Determine if we're running in a container."""
+    with open('/proc/1/sched', 'r') as sched:
+        data = sched.readline()
+
+    checks = [
+        data.split()[0] not in ('systemd', 'init',),
+        os.path.exists('/.dockerenv'),
+        os.path.exists('/.dockerinit'),
```
fcami: /.dockerenv and /.dockerinit are deprecated and will probably go away in the future. I'd rather not depend on these.

fcami: It might be better to check if the first line of /proc/1/cgroup does not contain "docker".

src/ipahealthcheck/system/filesystemspace.py
```diff
@@ -2,13 +2,29 @@
...
 from ipahealthcheck.core import constants
 
 
+def in_container():
+    """Determine if we're running in a container."""
+    with open('/proc/1/sched', 'r') as sched:
+        data = sched.readline()
+
+    checks = [
+        data.split()[0] not in ('systemd', 'init',),
```
fcami: I'm not sure whether that would produce a false negative in a systemd container (which I know nothing about atm).

rcritten: I don't know if IPA runs in it, that would I think be our driving interest.

fcami: Indeed. And that line takes care of both docker and podman/runc too.

fcami:
The container detection code should work on:

- standalone docker
- standalone podman/runc
- k8s + cri-o

The rest LGTM. ACK.

Issue #83, opened 2019-11-07
add check that sssd.conf on IPA master uses ipa provider for the IPA domain


abbra:
When checking `sssd.conf` on IPA masters, we need to make sure that provider for the primary IPA domain must be `ipa`, not anything else (like `ad`). Finding this through sosreport for environments where you aren't allowed to pull out data is tough.

rcritten:
To be clear you want to enforce that in the domain section for the IPA domain id_provider == 'ipa'?
What about the other options there? auth_provider, chpass_provider, and access_provider?

abbra:
Right. I think those other parameters need to be enforced as well.

PR #84, opened 2019-11-07
Ensure the IPA sssd domain has all providers configured as ipa

rcritten:
Debugging this remotely is very difficult so proactively look
for it.

https://github.com/freeipa/freeipa-healthcheck/issues/83

### DIFF
--- a/src/ipahealthcheck/ipa/trust.py
+++ b/src/ipahealthcheck/ipa/trust.py
@@ -196,6 +196,62 @@ class IPATrustDomainsCheck(IPAPlugin):
 
 
 @registry
+class IPADomainCheck(IPAPlugin):
+    """
+    Check that the IPA domain provider is configured to use ipa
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/ipa/trust.py
```diff
@@ -195,6 +195,66 @@ def check(self):
...
 
 
+@registry
+class IPADomainCheck(IPAPlugin):
+    """
+    Check that the IPA domain provider is configured to use ipa
+    """
+    @duration
+    def check(self):
+        if not self.registry.trust_agent:
```
abbra: I think this test should actually be run on any IPA master, not just trust agents.

rcritten: Sure, dropped the restriction.

abbra:
LGTM

Issue #86, opened 2019-11-19
'ipa-healthcheck --failures-only' when run on newly installed ipa-server lists ERROR for IPACertTracking check


rcritten:
Cloned from https://bugzilla.redhat.com/show_bug.cgi?id=1774032

'ipa-healthcheck --failures-only' when run on newly installed ipa-server lists ERROR for IPACertTracking check

```
# ipa-healthcheck --failures-only
[
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertTracking",
    "result": "ERROR",
    "uuid": "1a21e9ab-6b88-40b1-a279-d5c4607e192a",
    "when": "20191119130346Z",
    "duration": "0.303834",
    "kw": {
      "msg": "Missing tracking for cert-database=/etc/pki/pki-tomcat/alias, cert-nickname=caSigningCert cert-pki-ca, ca-name=dogtag-ipa-ca-renew-agent, cert-presave-command=/usr/libexec/ipa/certmonger/stop_pkicad, cert-postsave-command=/usr/libexec/ipa/certmonger/renew_ca_cert \"caSigningCert cert-pki-ca\", template-profile=None"
    }
  },
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertTracking",
    "result": "WARNING",
   "uuid": "486c5905-b00b-4b74-9250-c4ef00f7665d",
    "when": "20191119130346Z",
    "duration": "0.425676",
    "kw": {
      "key": "20191118100649",
      "msg": "Unknown certmonger id 20191118100649"
    }
  }
]
```

tscherf:
Since the patch is already merged, can we close this ticket?

rcritten:
Yes, it's done.

PR #87, opened 2019-11-19
Don't hardcode the set of tracked certificates

rcritten:
The set of certs to track is hardcoded in healthcheck and there are
two problems in this:

1. The caSigningCert is duplicated, one with an empty template-profile
   and one with the caCACert profile.
2. There is an error in the second which adds the caCACert such that
   it will never get added.

Drop this duplicated code and use the same call to retrieve the set
of CA and KRA certs as the upgrade code.

https://github.com/freeipa/freeipa-healthcheck/issues/86

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -3,6 +3,7 @@
 #
 
 from datetime import datetime, timezone, timedelta
+import itertools
 import logging
 import os
 import tempfile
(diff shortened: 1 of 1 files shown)


fcami:
Diff LGTM, it fixes the original issue on f31 + FreeIPA HEAD. ACK

tscherf:
Since the patch is already merged, can we close this ticket?

PR #95, opened 2019-12-02
Abstract plugin execution into separate class

rcritten:
Abstraction execution into a separate class so other packages can leverage the core processing of healthcheck without importing all the dependencies of ipa.

Only ipahealthcheck.ipa.core is required to execute generic tests.

A minimalistic project would start with:

```
from ipahealthcheck.core.core import RunChecks

def main():
    checks = RunChecks(['my.registry'],
             '/etc/my/healthcheck.conf')
    sys.exit(checks.run_healthcheck())
```


SilleBille:
This works perfectly in `pki-healthcheck` tool. https://github.com/dogtagpki/pki/pull/301 is currently blocked by this PR.

rcritten:
Added ack from previous comment.

Issue #58, opened 2019-07-29
Checking permissions


fcami:
Currently healthcheck only logs a warning if a file's permissions are not equal to the expected perms.
It should be WARNING if a file's permissions are more permissive and an ERROR if they are less permissive than expected.

PR #98, opened 2019-12-09
Add an additional test on POSIX permissions

dawud:
As per #58:

Currently healthcheck only logs a warning if a file's permissions are
not equal to the expected perms. It should be WARNING if a file's
permissions are more permissive and an ERROR if they are less permissive
than expected.


rcritten:
This looks good, can you squash them into a single commit?

rcritten:
Looks good, thanks.

Issue #75, opened 2019-09-25
Add exception for FileSystemChecks when running in a container


rcritten:
Not all checks may be applicable to a container so the check should be able to detect when it is running in a container and decide to skip one or more filesystems.

See https://github.com/freeipa/freeipa-container/issues/291

fcami:
See also https://github.com/freeipa/freeipa-container/issues/291#issuecomment-535177638

PR #99, opened 2019-12-09
Add additional test for container detection

dawud:
When running from an ubi8-init image with podman, the existing tests
implemented in 8c75bdf5 will not detect the container. Add a new test
checking for `libpod` in `/proc/self/cgroup`

Example:

```bash
$ podman run --name ubi8-init -d registry.access.redhat.com/ubi8-init

$ podman exec ubi8-init cat /proc/1/sched | head -1
systemd (1, #threads: 1)

$ podman exec ubi8-init cat /proc/self/cgroup
i0::/user.slice/user-1001.slice/user@1001.service/user.slice/libpod-c08663d4addf4722bbc8d708f5a2a131e565d933d23594fd2b22f98e7d72cc45.scope/crun-exec
```

Reference: #75

### DIFF
--- a/src/ipahealthcheck/system/filesystemspace.py
+++ b/src/ipahealthcheck/system/filesystemspace.py
@@ -13,10 +13,14 @@ from ipahealthcheck.core import constants
 def in_container():
     """Determine if we're running in a container."""
     with open('/proc/1/sched', 'r') as sched:
-        data = sched.readline()
+        data_sched = sched.readline()
+
+    with open('/proc/self/cgroup', 'r') as cgroup:
+        data_cgroup = cgroup.readline()
 
     checks = [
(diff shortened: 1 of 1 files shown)


rcritten:
As discussed offline this should probably be something like:

result = ipautil.run(['/bin/systemd-detect-virt', '-c'], raiseonerr=False,
                     capture_output=True)
if result.returncode != 0:
    # fail maybe? Ignore? Log?
return 'container' in result.output.strip()

rcritten:
I tested systemd-detect-virt -c in a F30 kvm image and ran docker run -it fedora:27 /bin/bash and systemd-detect-virt returns kvm, with -c returns none. So I'm going to stick with the current detection code.

So this change looks good to me. Thanks for the detailed reproducer.

PR #101, opened 2019-12-10
Clarifies the predefined keys of the kw argument of the Result class

antonionc:
Rewording of the README.md to reflect the code in core/plugin.py Result class with regards to the way to pass the predefined keys of  `kw` in the constructor.


rcritten:
Looks good, thanks.

PR #102, opened 2019-12-10
Patch to allow plugins to read config dict

SilleBille:
This patch allows the plugins to use the values parsed from the
`healthcheck.conf` file while initializing. This improves the way
a plugin can use the values defined in the conf

Thanks @rcritten for the patch.

**Note:** Currently, the dictionary created is compiled into a **single dictionary**. There is no way to differentiate values from different sections

`Signed-off-by: Dinesh Prasanth M K <dmoluguw@redhat.com>`

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -64,7 +64,7 @@ def source_or_check_matches(plugin, source, check):
     return True
 
 
-def run_service_plugins(plugins, config, source, check):
+def run_service_plugins(plugins, source, check):
     """Execute plugins with the base class of ServiceCheck
 
        This is a specialized check to use systemd to determine
@@ -90,7 +90,7 @@ def run_service_plugins(plugins, config, source, check):
     return results, set(available)
 
 
-def run_plugins(plugins, config, available, source, check):
+def run_plugins(plugins, available, source, check):
     """Execute plugins without the base class of ServiceCheck
 
        These are the remaining, non-service checking checks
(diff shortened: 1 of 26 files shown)


src/ipahealthcheck/core/main.py
```diff
@@ -220,7 +219,7 @@ def main():
 
     for name, registry in find_registries().items():
         try:
-            registry.initialize(framework)
+            registry.initialize(framework, config)
         except Exception as e:
             print("Unable to initialize %s: %s" % (name, e))
             sys.exit(1)
```
rcritten: It also needs this:

diff --git a/src/ipahealthcheck/core/main.py b/src/ipahealthcheck/core/main.py
index 3698c9e..05fac35 100644
--- a/src/ipahealthcheck/core/main.py
+++ b/src/ipahealthcheck/core/main.py
@@ -247,10 +247,10 @@ def main():
         if options.source:
             results = limit_results(results, options.source, options.check)
     else:
-        results, available = run_service_plugins(plugins, config,
+        results, available = run_service_plugins(plugins,
                                                  options.source,
                                                  options.check)
-        results.extend(run_plugins(plugins, config, available,
+        results.extend(run_plugins(plugins, available,
                                    options.source, options.check))
 
     if options.source and len(results.results) == 0:

SilleBille: Done

rcritten:
Can you rebase this?

And this change moves the config object from the plugins to the registry, can you update the commit message to reflect that?

rcritten:
Change looks ok, needs a rebase.

SilleBille:
Done

SilleBille:
@rcritten thank you! Also, check my note in this PR's description. Should I file a issue for it?

rcritten:
Feel free to file an RFE

Issue #107, opened 2020-01-02
Don't consider missing config file a hard fail


rcritten:
Currently if the configuration file cannot be read the script fails hard. Make it a warning message and continue with defaults.

PR #108, opened 2020-01-02
Don't consider a missing configuration file to be fatal

rcritten:
If no config file then log a warning and use all the built-in
defaults.

https://github.com/freeipa/freeipa-healthcheck/issues/107


fcami:
LGTM, works as expected and no regression noted with output-file. ACK.

PR #109, opened 2020-01-10
Add Directory Server Healthchecks from lib389

marcus2376:
Add all the current DS healthchecks fom lib389 into IDM healthcheck.
Also removed the current conflict check as this is now covered by
the lib389 Replica checks.


rcritten:
https://github.com/freeipa/freeipa-healthcheck/pull/111 is probably going to be necessary to get this to pass. It drops testing F29 which is EOL anyway.

rcritten:
If you rebase then it should pass CI. I removed the F-27 test.

rcritten:
This is looking good. What is the plan for the TLS 1.0 false positive?

I'm testing with 389-ds-base-1.4.1.13-1.fc30

marcus2376:
> This is looking good. What is the plan for the TLS 1.0 false positive?
> 
> I'm testing with 389-ds-base-1.4.1.13-1.fc30

Well it's not a false positive, this is what NSS returns as the minimum version that it allows(on F30).  I heard in newer versions of NSS the minimum version was raised to 1.2 - so this will kind of fix itself.  Other than that we have talked about hard-coding the minimum to 1.2, but this is still being discussed.

rcritten:
Ok. I'm not keen on the out-of-the-box experience resulting in an error I guess.

I'll wait to see what the result of the TLS 1.2-as-default discussion is.

marcus2376:
> Ok. I'm not keen on the out-of-the-box experience resulting in an error I guess.
> 
> I'll wait to see what the result of the TLS 1.2-as-default discussion is.

This is where we left it:

https://pagure.io/389-ds-base/issue/49395

So in RHEL 8 this is supposedly not an issue.  This is more of a upstream issue.  I don't really want to hard code the minimum when we know it's not an issue moving forward.  Let me confirm how RHEL 8 and F31 behave...

marcus2376:
Hmm in rawhide its still reporting TLS1.0 as the min, need to check a RHEL 8 machine next...

marcus2376:
RHEL 8.1 does indeed default to TLS1.2 for the minimum.  Will discuss with the team on how to proceed...

abbra:
@marcus2376 this is on purpose. Fedora refused to tighten DEFAULT crypto policy to start at TLS 1.2. It means 389-ds works just fine by picking up crypto-policy settings from the system.

rcritten:
Makes me wonder if IPA should be enforcing TLS 1.2+.

simo5:
On Thu, 2020-01-16 at 16:34 +0000, Rob Crittenden wrote:
> Makes me wonder if IPA should be enforcing TLS 1.2+.

You can, but I would rather use the system default, esp where crypto
policies are available (Fedara 26+, RHEL 8+, CentOs 8+ ..)

Simo.

-- 
Simo Sorce
RHEL Crypto Team
Red Hat, Inc

rcritten:
Agreed. My point is we are a security product. When running our own checker against our security product the first error that pops up in a default install should not be "you are insecure".

simo5:
> Agreed. My point is we are a security product. When running our own checker against our security product the first error that pops up in a default install should not be "you are insecure".

If defaults are used then this would be true, I read your former comment as enforcing in all cases that IPA wouldn't use anything but TLS 1.2+ It should be the default for IPA/Idm on RHEL8 that only TLS1.2+ is allowed by default but not on older RHEL or if policy has been set to LEGACY

marcus2376:
I am proposing in DS to set the default minimum to 1.2.  Downstream this will only land in 8.2 or 8.3, it's upstream where it would be different.  We can however set the TLS min/max in the DS config.  So we wouldn't be forcing anyone to use 1.2 as the minimum, but out-of-the-box we would default to it.

rcritten:
Working without warning with 389-ds-base-1.4.1.14-1.fc30+

PR #111, opened 2020-01-10
Remove testing of Fedora 29, it is EOL

rcritten:
It was tested originally because it tested against the IPA 4.7.x
branch.


rcritten:
self-ack since this is so trivial.

PR #114, opened 2020-01-16
[Minor] Move ipa-healthcheck man category to 8

SilleBille:
The `ipa-healthcheck` command needs to be executed as a superuser since it
requires special privileges. Category 1 man pages are used for regular
user-commands, while Category 8 is used for superuser commands. Hence,
moving the ipa-healthcheck to the right category.

Ref: https://linux.die.net/man/8/intro

`Signed-off-by: Dinesh Prasanth M K <dmoluguw@redhat.com>`


rcritten:
Thanks, looks good.

PR #117, opened 2020-01-27
Add cluster checking capability

rcritten:
This is an experiment to build cluster-wide analysis on top of the healthcheck framework.

Cluster checking is at its core reading in all the healthcheck logs from all the IPA masters and running plugins to do analysis. That is so close to healthcheck, minus the log reading, that it seems natural to not re-invent the wheel.

I included a simple cluster checker for validation, crlmanager enforcement. There can be only one. The data returned is still a bit TBD, I need to work with Thomas how we should report error conditions in a way he can tell. Right now it is included in the same output as a successful report.

This has a backwards-compatibility breaking feature: adding options to the registry initialization.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,7 +4,7 @@ from setuptools import find_packages, setup
 setup(
     name='ipahealthcheck',
     version='0.5',
-    namespace_packages=['ipahealthcheck'],
+    namespace_packages=['ipahealthcheck', 'ipaclustercheck'],
     package_dir={'': 'src'},
     # packages=find_packages(where='src'),
     packages=[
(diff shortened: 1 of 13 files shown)


fcami:
No real remark on the code additions and changes except LGTM.

I tried on a healthy cluster:
```
# ipa-clustercheck --directory hc
config file /etc/ipa/clustercheck.conf does not exist, using defaults
[
  {
    "crlmanager": "ipa0.laptop.example.org"
  }
]
# echo $?
0
```

and on an unhealthy cluster:
```
# ipa-clustercheck --directory hc2
config file /etc/ipa/clustercheck.conf does not exist, using defaults
[
  {
    "crlmanager": "Multiple CRL Managers defined: ipa0.laptop.example.org,ipa1.laptop.example.org"
  }
]
# echo $?
1
```

ACK from me.

PR #118, opened 2020-02-05
Collect the entire list of masters in the meta plugin

rcritten:
This is so the total known masters is known when doing cluster
evaluation. A master may not have provided its healthcheck log
so it will be unknonw to the cluster.


fcami:
Works as expected, LGTM.
I can't think of a better way to do it. ACK.

PR #119, opened 2020-02-07
Cluster RUV checking

rcritten:
Add Cluster RUV checking. It does a similar test that ipa-replica-manage clean-dangling-ruv does.

It also:
- adds new IPA Meta to provide the full list of known masters. This isn't yet compared internally. We'll need to decide whether to take the longest version, throw up if things don't match, etc.
- test for cluster RUV. The data is a bit of a downer, it may need to be moved to discrete files for better readability if it grows any more.
- A few small tweaks to the general cluster reporting framework.

### DIFF
--- a/tests/clusterdata.py
+++ b/tests/clusterdata.py
@@ -1,0 +1,895 @@
+#
+# Copyright (C) 2019 FreeIPA Contributors see COPYING for license
+#
+
+
+ONE_MASTER = {
+    'ipa.ipa.example': [
+        {
+            "source": "ipahealthcheck.meta.core",
+            "check": "MetaCheck",
+            "result": "SUCCESS",
+            "kw": {
+              "fqdn": "ipa.ipa.example",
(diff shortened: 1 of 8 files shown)


tests/clusterdata.py
```diff
@@ -0,0 +1,895 @@
...
+            "source": "ipahealthcheck.ds.ruv",
+            "check": "RUVCheck",
+            "result": "SUCCESS",
+            "kw": {
+              "key": "o=ipaca",
+              "ruv": "6"
+            }
+        },
+        {
+            "source": "ipahealthcheck.ds.ruv",
```
fcami: I might be mistaken but the source and data seem mismatched (ds vs cs).

fcami:
ACK. I found it confusing that ds.ruv would be the data source that reports csruvs.
I don't think it matters as in the end, all that matters is whether there are dangling RUVs or not.

PR #121, opened 2020-03-02
Fix man page to be consistent

abbra:
- --output-type description didn't specify possible types
- examples used non-existing option --output-format
- use .nf/.fi for pre-formatted text

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1809215

### DIFF
--- a/man/man8/ipa-healthcheck.8
+++ b/man/man8/ipa-healthcheck.8
@@ -37,7 +37,7 @@ Execute one or more checks within this given source.
 Execute this particular check within a source. A source must be supplied as well with this option.
 .TP
 \fB\-\-output\-type\fR=\fITYPE\fR
-Set the output type. The default is JSON.
+Set the output type. Supported variants are \fBhuman\fR and \fBjson\fR. The default is \fBjson\fR.
 .TP
 \fB\-\-failures\-only\fR
 Exclude SUCCESS results on output.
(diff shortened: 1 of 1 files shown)


man/man8/ipa-healthcheck.8
```diff
@@ -37,7 +37,7 @@ Execute one or more checks within this given source.
 Execute this particular check within a source. A source must be supplied as well with this option.
 .TP
 \fB\-\-output\-type\fR=\fITYPE\fR
-Set the output type. The default is JSON.
+Set the output type. Supported variants are \fBhuman\fR and \fBjson\fR. The default is JSON.
```
rcritten: The default should be lower-case json, not JSON. I just double-checked this. It is case sensitive.

abbra: Yep, this was the original sentence. I fixed it now.

man/man8/ipa-healthcheck.8
```diff
@@ -70,28 +70,28 @@ Take as input a JSON results output and convert it to a more human\-readable for
...
+\&# ipa\-healthcheck \-\-failures\-only
+.fi
 .PP
 Display in human\-readable output a previous report:
 .PP
-.Vb 2
-\& ipa\-healthcheck \-\-output\-format human \-\-input\-file \e
+.nf 2
+\&# ipa\-healthcheck \-\-output\-type human \-\-input\-file \e
 \&        /var/log/ipa/check.json
```
rcritten: Can you change one more thing, the log location in the sample to /var/log/ipa/healthcheck/healthcheck.log ? This is the default location that the systemd timer logs to

abbra: Done.

rcritten:
Ack, thanks for the patch

PR #128, opened 2020-06-09
enhance --source to specify source namespace

frasertweedale:
As a FreeIPA developer I want ipa-healthcheck to be able to execute
all sources in a given namespace, excluding others, so that
addition/removal of checks by other projects (e.g. Dogtag) does not
affect test results.

Enhance the --source option such that its argument is treated as a
namespace.


rcritten:
Works great, thanks for the patch.

PR #129, opened 2020-06-16
Don't completely fail if a registry fails to initialize

rcritten:
Log the error if registry initialization fails. This will
suppress a bad plugin from preventing healthcheck from
executing at all.

Currently this will only be reported on stdout so it won't
be immediately obvious that an entire registry is skipped.


fcami:
ACK

PR #131, opened 2020-06-23
Support the older registry initialization API for another release

rcritten:
Initially this would be caught and the external plugin just ignored.
Add a deprecation message instead so they have a chance to catch
up and fix their initialization.

This will allow the current pki healthcheck to execute.

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -6,6 +6,7 @@ import argparse
 import json
 import logging
 import pkg_resources
+import warnings
 
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


fcami:
LGTM. @SilleBille this is all yours now.

src/ipahealthcheck/core/core.py
```diff
@@ -264,8 +265,13 @@ def run_healthcheck(self):
...
             except Exception as e:
-                logger.error("Unable to initialize %s: %s" % (name, e))
-                continue
+                warnings.warn("Trying deprecated initialization API: %s" % e,
+                              DeprecationWarning)
+                try:
+                    registry.initialize(framework, config)
+                except Exception as e:
+                    logger.error("Unable to initialize %s: %s" % (name, e))
+                    continue
```
fcami, line 274: ACK

SilleBille: oh, sorry missed to comment this PR. Looks good! Thanks!

Issue #133, opened 2020-06-25
Config object has a recursive attribute


rcritten:
@SilleBille found an issue with the Config object: it's recursive.

```
    from ipahealthcheck.core.config import read_config
     
    '''
    # cat /etc/pki/healthcheck.conf
    [default]
    instance=pki-tomcat
     
    [dogtag]
    ca_instance=pki-ca
    kra_instance=pki-kra
    ocsp_instance=pki-ocsp
    '''
     
    config = read_config('/etc/pki/healthcheck.conf')
     
    print(config._Config__d['_Config__d'])
    print(config._Config__d['_Config__d']['_Config__d'])
    print(config._Config__d['_Config__d']['_Config__d']['_Config__d'])
    print(config._Config__d['_Config__d']['_Config__d']['_Config__d']['_Config__d'])
```

PR #134, opened 2020-06-25
Fix recursion in Config object

rcritten:
Fix recursion in the Config object
    
The data dictionary was incorrectly recursive, ala
config._Config__d['_Config__d']['_Config__d']['_Config__d'])
    
https://github.com/freeipa/freeipa-healthcheck/issues/133

### DIFF
--- a/src/ipahealthcheck/core/config.py
+++ b/src/ipahealthcheck/core/config.py
@@ -30,7 +30,7 @@ class Config:
     """
 
     def __init__(self):
-        self.__d = dict()
+        object.__setattr__(self, '_Config__d', {})
 
     def __setattr__(self, key, value):
         """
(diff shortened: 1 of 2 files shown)


SilleBille:
Tested this on my local with my test script from #133 . I no longer see the recursive attribute but i do see the value **twice**.

Question: How does the `_Config__d` variable get initialized, even though we don't declare it explicitly? 

````
>>> print(vars(config))
{'_Config__d': {'cert_expiration_days': '675'}, 'cert_expiration_days': '675'}
````

rcritten:
__d is the dictionary that the config object uses, Config is the class. I'm assuing that _Config_d is what python calls this internally. I guess it shows up in vars() because it is included but this is a class variables.

If you instead do something like:

>>> x = list(Config)
>>> print(x)
['cert_expiration_days', 'instance']

Issue #120, opened 2020-02-14
RFE: check that the admin user exists


fcami:
Additionally check (in trust.py) that it has a proper SID.

PR #139, opened 2020-07-13
Verify on trusts that the IPA admin users exists with right SID

rcritten:
The admin user is required for trusts and must have the Domain
Admin RID 500.

https://github.com/freeipa/freeipa-healthcheck/issues/120


fcami:
Ack'ed.

Issue #105, opened 2020-01-02
Report FIPS status in meta plugin


rcritten:
It could be useful to know if FIPS is enabled, report its status in the meta plugin.

Can either read /proc/sys/crypto/fips_enabled or use fips-mode-setup --check

PR #141, opened 2020-07-15
Report on FIPS status: enabled, disabled, inconsistent, etc.

rcritten:
Report on FIPS status: enabled, disabled, inconsistent, etc.
    
If fips-mode-setup is installed then we can check the status
otherwise report the missing binary.
    
The script provides 3 possible return values:
0 enabled
1 inconsistent
2 disabled
    
These are handled along with a catch-all "unknown"
   
https://github.com/freeipa/freeipa-healthcheck/issues/105

### DIFF
--- a/src/ipahealthcheck/meta/core.py
+++ b/src/ipahealthcheck/meta/core.py
@@ -12,6 +12,9 @@ from ipapython import ipautil
 from ipapython.version import VERSION, API_VERSION
 from ipaplatform.paths import paths
 
+if 'FIPS_MODE_SETUP' not in dir(paths):
+    paths.FIPS_MODE_SETUP = '/usr/bin/fips-mode-setup'
+
 logger = logging.getLogger()
 
 
--- a/src/ipahealthcheck/meta/core.py
+++ b/src/ipahealthcheck/meta/core.py
@@ -2,18 +2,52 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 
+import logging
+import os
 import socket
 from ipahealthcheck.core import constants
 from ipahealthcheck.core.plugin import Result, duration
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/meta/core.py
```diff
@@ -2,18 +2,49 @@
...
+
 
 @registry
 class MetaCheck(Plugin):
     @duration
     def check(self):
-        yield Result(self, constants.SUCCESS,
+
+        rval = constants.SUCCESS
+        if not os.path.exists('/usr/bin/fips-mode-setup'):
```
fcami, line 23: Does it make sense to put that into ipaplatform.paths instead of hardcoding the path here?

rcritten: If IPA wants to add this then we could consume it pretty easily. I can open a PR against IPA but it would require an upstream release before we could utilize it.

fcami: Can you open a PR against IPA and file an issue against freeipa-healthcheck so that we don't forget?

rcritten: freeIPA upstream issue: https://pagure.io/freeipa/issue/8429 and PR https://github.com/freeipa/freeipa/pull/4942

fcami: https://github.com/freeipa/freeipa/pull/4942 is merged. Should we adjust things here or just file an issue against freeipa-healthcheck and do the change later (for backport reasons)?

rcritten: We need to wait until the next upstream release and then we can do it and tie the two together.

rcritten:
Based on previous discussions I'm going to go ahead and merge this.

Issue #144, opened 2020-07-28
pki-healthcheck fails with 'Namespace' object has no attribute 'failures_only' with 0.6


rcritten:
Reported against Fedora 31 with freeipa-healthcheck-0.6-1 and pki-server-10.9.0-0.2.fc31

```
# pki-healthcheck
Traceback (most recent call last):
  File "/usr/sbin/pki-healthcheck", line 11, in <module>
    load_entry_point('pkihealthcheck==0.1', 'console_scripts', 'pki-healthcheck')()
  File "/usr/lib/python3.8/site-packages/pki/server/healthcheck/core/main.py", line 31, in main
    sys.exit(checks.run_healthcheck())
  File "/usr/lib/python3.8/site-packages/ipahealthcheck/core/core.py", line 280, in run_healthcheck
    output = out(options)
  File "/usr/lib/python3.8/site-packages/ipahealthcheck/core/output.py", line 102, in __init__
    super(JSON, self).__init__(options)
  File "/usr/lib/python3.8/site-packages/ipahealthcheck/core/output.py", line 40, in __init__
    self.failures_only = options.failures_only
AttributeError: 'Namespace' object has no attribute 'failures_only'
```

PR #145, opened 2020-07-28
Allow consuming projects to not use all available options

rcritten:
Some consumers of the framework may not be interested in the
filtering options. Don't fail if some or all of those are not
defined as options. Set logical defaults in this case.

https://github.com/freeipa/freeipa-healthcheck/issues/144

### DIFF
--- a/tests/test_init.py
+++ b/tests/test_init.py
@@ -1,0 +1,28 @@
+#
+# Copyright (C) 2020 FreeIPA Contributors see COPYING for license
+#
+
+import argparse
+
+from ipahealthcheck.core.output import output_registry
+
+
+class RunChecks:
+    def run_healthcheck(self):
+        options = argparse.Namespace(check=None, debug=False, indent=2,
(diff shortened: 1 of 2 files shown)


fcami:
@rcritten this LGTM. Do we want a test to make sure we don't change these in the future?

rcritten:
Yes, that would be ideal. Let me see if we can do something similar to the way pki-healthcheck imports things. It's the invocation that might be the tougher part since dependencies might not be satisfied.

tests/test_init.py
```diff
@@ -0,0 +1,28 @@
...
+                out(options)
+                break
+
+
+def test_run_healthcheck():
+    """
+    Test typical initialization in run_healthcheck (based ok pki-healthcheck)
+    """
+    run = RunChecks()
+    run.run_healthcheck()
```
fcami, line 28: LGTM. ACK, even.

fcami, line 28: LGTM. ACK, even.

Issue #146, opened 2020-08-04
error in custom value of cert_expiration_days


stanislavlevin:
```
cat /etc/ipahealthcheck/ipahealthcheck.conf
[default]
cert_expiration_days=10
```

```
ipa-healthcheck
...
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertmongerExpirationCheck",
    "result": "CRITICAL",
    "uuid": "7b7a66be-949e-4ff7-a5a8-161c0036eee2",
    "when": "20200804092812Z",
    "duration": "0.007589",
    "kw": {
      "exception": "'<' not supported between instances of 'int' and 'str'"
    }
  },
...
```

```
  /usr/bin/ipa-healthcheck(11)<module>()
-> load_entry_point('ipahealthcheck==0.6', 'console_scripts', 'ipa-healthcheck')()
  /usr/lib64/python3/site-packages/ipahealthcheck/core/main.py(44)main()
-> sys.exit(ipachecks.run_healthcheck())
  /usr/lib64/python3/site-packages/ipahealthcheck/core/core.py(304)run_healthcheck()
-> options.source, options.check))
  /usr/lib64/python3/site-packages/ipahealthcheck/core/core.py(122)run_plugins()
-> for result in run_plugin(plugin, available):
  /usr/lib64/python3/site-packages/ipahealthcheck/core/core.py(44)run_plugin()
-> for result in plugin.check():
  /usr/lib64/python3/site-packages/ipahealthcheck/core/plugin.py(18)wrapper()
-> for result in f(*args, **kwds):
> /usr/lib64/python3/site-packages/ipahealthcheck/ipa/certs.py(211)check()
-> if diff < self.config.cert_expiration_days:
(Pdb) vars(self.config)
{'_Config__d': {'cert_expiration_days': '10'}, 'cert_expiration_days': '10'}
```

Explicit type conversion is required.

PR #147, opened 2020-08-04
Explicitly convert cert_expiration_days to int

stanislavlevin:
According to docs \[0\]:
> Supported Datatypes
Config parsers do not guess datatypes of values in configuration files,
always storing them internally as strings. This means that if you need
other datatypes, you should convert on your own.

\[0\]: https://docs.python.org/3/library/configparser.html#supported-datatypes

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/146


rcritten:
Checks out. This falls into the category of "how did I never notice this". Thanks for the patch. ACK.

PR #153, opened 2020-10-27
Run Travis CI against Fedora 32 and 33 instead of 30 and 31.

rcritten:


rcritten:
Moved into PR 155 to exercise that code change.

PR #155, opened 2020-10-27
Mock the ipa-ca resolution for the DNS system records check

rcritten:
IPA uses the resolver to look up the IP addresses for those
servers with a CA to ensure that the ipa-ca SAN will work
properly.

Mock this call and return our own rrsets.


rcritten:
I updated this PR to include the Travis changes to run in F32/33 instead of F30/31.

rcritten:
CI is passing, merging.

PR #156, opened 2020-10-30
Support an additional trust type, don't rely on strings for type

rcritten:
ipaserver/dcerpc_common.py defines the possible trust types:

_trust_type_dict = {
 1: _('Non-Active Directory domain'),
 2: _('Active Directory domain'),
 3: _('RFC4120-compliant Kerberos realm'),
 10: _('Non-transitive external trust to a domain in '
       'another Active Directory forest'),
 11: _('Non-transitive external trust to an RFC4120-'
       'compliant Kerberos realm')
}

In practice, only '2' and '10' can appear as trust types right now.
The other three describe possible options for the attributes on AD
side. When IPA-IPA trust will be added, it will be seen as '2'.

Since this is a translated value use ipanttrusttype instead of
trusttype.

https://bugzilla.redhat.com/show_bug.cgi?id=1891505

### DIFF
--- a/src/ipahealthcheck/ipa/trust.py
+++ b/src/ipahealthcheck/ipa/trust.py
@@ -15,6 +15,10 @@ from ipalib import api
 from ipaplatform.paths import paths
 from ipapython import ipautil
 from ipapython.dn import DN
+from ipaserver.dcerpc_common import (
+    trust_type_string,
+    _trust_type_dict_unknown
+)
 
 try:
     import pysss_nss_idmap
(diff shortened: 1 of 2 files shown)


abbra:
You only took `ipanttrusttype` but the dict keys are built from a combination of `ipanttrusttype` and `ipanttrustattributes`, so you'd never match '10' . Please use the formula from `trust_type_string()` to check the types.

Since this code exists in any IPA server installation, may be you can simply ask for `trust_type_string(ipanttrusttype, ipanttrustattributes)` and check if resulted value is not `_trust_type_dict_unknown`?

abbra:
While you are at this code, could you please make sure flatnames of those AD domains in the tests are unique?

rcritten:
Done. I made the SID unique as well.

rcritten:
Alexander gave me an ack online, merging.

Issue #159, opened 2020-11-06
[self-tests] test_expiration_warning fails since Nov 6 2020


stanislavlevin:
Since Nov 6 2020 `test_expiration_warning` fails with:
```
=================================== FAILURES ===================================
____________________ TestExpiration.test_expiration_warning ____________________

self = <test_ipa_expiration.TestExpiration testMethod=test_expiration_warning>

    def test_expiration_warning(self):
        warning = datetime.now(timezone.utc) + timedelta(days=20)
        replaceme = {
            'nickname': '7777',
            'cert-file': paths.RA_AGENT_PEM,
            'key-file': paths.RA_AGENT_KEY,
            'ca-name': 'dogtag-ipa-ca-renew-agent',
            'not-valid-after': int(warning.timestamp()),
        }
    
        set_requests(remove=0, add=replaceme)
    
        framework = object()
        registry.initialize(framework, config.Config)
        f = IPACertmongerExpirationCheck(registry)
    
        f.config.cert_expiration_days = '30'
        self.results = capture_results(f)
    
        assert len(self.results) == 2
    
        result = self.results.results[0]
>       assert result.result == constants.SUCCESS
E       AssertionError: assert 10 == 0
E        +  where 10 = ipahealthcheck.ipa.certs.IPACertmongerExpirationCheck({'key': '5678', 'expiration_date': '20201205214850Z', 'days': 29... expires in {days} days. certmonger should renew this automatically. Watch the status withgetcert list -i {key}.'}): 10.result
E        +  and   0 = constants.SUCCESS

tests/test_ipa_expiration.py:76: AssertionError
=========================== short test summary info ============================
FAILED tests/test_ipa_expiration.py::TestExpiration::test_expiration_warning
======================== 1 failed, 110 passed in 4.17s =========================
```

PR #160, opened 2020-11-06
tests: Generate a proper `not-valid-after` field

stanislavlevin:
Some tests assume that the mocked certificate will be valid in N
days from now(). There was a hardcoded `not-valid-after` value
which pointed to 20201205214850Z. So, from Nov 06 2020 the assertion
20201205214850Z - now() < cert_expiration_days(30days) fails.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/159

### DIFF
--- a/tests/mock_certmonger.py
+++ b/tests/mock_certmonger.py
@@ -3,12 +3,15 @@
 #
 
 import copy
+from datetime import datetime, timedelta, timezone
 
 from ipaplatform.paths import paths
 
 # Fake certmonger tracked request list. This is similar but can be
 # distinct from the value from the overrident get_defaults() method.
 template = paths.CERTMONGER_COMMAND_TEMPLATE
+
+CERT_EXPIRATION_DAYS = 30
 
 pristine_cm_requests = [
     {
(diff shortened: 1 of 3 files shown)


rcritten:
I am not a fan of black.

stanislavlevin:
As you wish :) just a moment.

tests/mock_certmonger.py
```diff
@@ -20,7 +23,11 @@
         'cert-storage': 'FILE',
         'cert-presave-command': template % 'renew_ra_cert_pre',
         'cert-postsave-command': template % 'renew_ra_cert',
-        'not-valid-after': 1024,
+        'not-valid-after': (
+            int(
+                datetime(1970, 1, 1, 0, 17, 4, tzinfo=timezone.utc).timestamp()
```
rcritten, line 28: A fancier way of doing it :-) I  think this was my way of forcing an guaranteed-expired cert but formalizing it is a good thing.

tests/mock_certmonger.py
```diff
@@ -30,7 +37,14 @@
         'template_profile': 'caIPAserviceCert',
         'cert-storage': 'FILE',
         'cert-postsave-command': template % 'restart_httpd',
-        'not-valid-after': 1607204930,
```
rcritten, line 33: Aaaarg. I'm sure I meant to go back and fix this at some point...

rcritten:
This looks good, thanks.

Issue #168, opened 2020-11-09
Deprecate Travis CI in favor of Github Actions


rcritten:
Travis is migrating its runners from .org to .com (https://docs.travis-ci.com/user/migrate/open-source-repository-migration)

It is claimed that open source projects will proceed as in the past but https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing suggests it will be non-free at some point.

In any case it is sometimes taking hours to execute CI because of the limited number of runners remaining on travis.org.

PR #169, opened 2020-11-09
Use GitHub actions for CI instead of Travis

rcritten:
Travis CI has been really slow since the migration. Use GitHub Actions instead for CI.

https://github.com/freeipa/freeipa-healthcheck/issues/168


.github/workflows/pipelines.yml
```diff
@@ -0,0 +1,73 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        fedora-release: [32, 33]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Lint with pylint fedora:${{ matrix.fedora-release }}
+      run: |
+        docker pull fedora:${{ matrix.fedora-release }}
```
SilleBille, line 44: Have you tried the service `container` option to avoid dealing with `docker` manually? It could cut down the code maintenance. My $0.02 😄 

Reference: https://github.com/dogtagpki/pki/blob/master/.github/workflows/ipa-tests.yml#L16 
Documentation: https://docs.github.com/en/free-pro-team@latest/actions/guides/about-service-containers

rcritten: I looked into that and it seemed more for running a service than executing the tests. Since healthcheck isn't a service I didn't think it really applied.

fcami (approved):
ACK from me

PR #170, opened 2020-11-09
Use GitHub Actions for CI

rcritten:
I tried enabling it via a fork but it seems like it needs to be enabled via a direct change


rcritten:
Closing this in favor of https://github.com/freeipa/freeipa-healthcheck/pull/169

Let's hope the actions stay enabled.

PR #171, opened 2020-11-17
Use trust-find and trustdomain-find to identify all AD trusts

rcritten:
Not all AD domains are visible to trust-find. For each trust
iterate over trustdomain-find <domain> to find the complete
list of domains.


abbra:
LGTM, simpler and more future-proof.

PR #173, opened 2020-12-03
Remove deprecated options from setup.py and tox.ini

rcritten:
Remove deprecated options from setup.py

Remove ‘pytest-runner’ from your ‘setup_requires’, preferably
removing the setup_requires option.

Remove ‘pytest’ and any other testing requirements from
‘tests_require’, preferably removing the setup_requires option.

We already use tox for basic environment setup.

https://github.com/freeipa/freeipa-healthcheck/issues/161

Also replace pep8 with pycodestyle.


fcami (approved):
ACK

PR #178, opened 2021-01-13
Use full import path for imports from tests directory

rcritten:
It recently started failing with:

Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_ipa_dna.py:5: in <module>
    from base import BaseTest
E   ModuleNotFoundError: No module named 'base'

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -2,8 +2,8 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 from unittest import mock, TestCase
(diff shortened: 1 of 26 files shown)


rcritten:
Closing this, it isn't necessary. I had created __init__.py in tests/ making it a module.

PR #179, opened 2021-01-14
In installation check be specific about server

rcritten:
Previously if IPA server was not configured the message was
"IPA is not configured." If ipa-healthcheck is run on a client
then this could be misleading.


src/ipahealthcheck/core/main.py
```diff
@@ -15,7 +15,7 @@
 class IPAChecks(RunChecks):
     def pre_check(self):
         if not is_ipa_configured():
-            print("IPA is not configured")
+            print("IPA server is not configured")
```
fcami, line 18: ACK

fcami (approved):

PR #182, opened 2021-01-14
Handle slight different in exception output in Python 3.10

rcritten:
Python 3.10 adds the class name to the exception so a
x = Result() failure will return:

Result.__init__() missing...

instead as with previous python versions:

__init__() missing...

https://bugzilla.redhat.com/show_bug.cgi?id=1915256


fcami (approved):
ACK

Issue #183, opened 2021-01-18
Expanded user/group per file is reporting group as a list, not a string


rcritten:
IPA test test_integration/test_ipahealthcheck.py::TestIpaHealthCheckFileCheck::test_tomcat_filecheck_bad_group is failing in 0.8 with:

```
self = <ipatests.test_integration.test_ipahealthcheck.TestIpaHealthCheckFileCheck object at 0x7f33b4d6df70>
modify_permissions = <function modify_permissions.<locals>._modify_permission at 0x7f33b549ddc0>

    def test_tomcat_filecheck_bad_group(self, modify_permissions):
        modify_permissions(self.master, path=paths.CA_CS_CFG_PATH,
                           group='root')
        returncode, data = run_healthcheck(
            self.master,
            "ipahealthcheck.ipa.files",
            "TomcatFileCheck",
            failures_only=True,
        )
        assert returncode == 1
        for check in data:
            assert check["result"] == "WARNING"
            assert check["kw"]["key"] == \
                '_var_lib_pki_pki-tomcat_conf_ca_CS.cfg_group'
            assert check["kw"]["type"] == 'group'
>           assert check["kw"]["expected"] == 'pkiuser'
E           AssertionError: assert ['pkiuser'] == 'pkiuser'
E             +['pkiuser']
E             -'pkiuser'
```

rcritten:
https://github.com/freeipa/freeipa-healthcheck/pull/184

rcritten:
Added as a patch to Fedora 32-34. 

Submitted to bodhi:

https://bodhi.fedoraproject.org/updates/FEDORA-2021-8f96bcecb5
https://bodhi.fedoraproject.org/updates/FEDORA-2021-7399dee084

PR #184, opened 2021-01-18
Failed expected group should be a string, not a list

rcritten:
This was caused by the code to allow multiple possible
file owner and group. The expected value for group wasn't
being converted into a string so was kept as a list.

Test updated to catch this situation.

https://github.com/freeipa/freeipa-healthcheck/issues/183


flo-renaud (approved):
@rcritten thanks for the PR, LGTM

PR #185, opened 2021-02-15
Return user-friendly message when no issues found

antoniotorresm:
Return user-friendly message instead of empty string
when no issues found and using the "human" output type.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1780062


rcritten:
LGTM

PR #188, opened 2021-02-23
Add check for IPA KRA Agent

antoniotorresm:
Add check to validate KRA Agent in case KRA is installed, including
checking for the KRA Agent LDAP entry.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1894781

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -724,6 +724,83 @@ class IPAOpenSSLChainValidation(IPAPlugin):
                         self, constants.SUCCESS, key=cert)
 
 
+def check_agent(plugin, base_dn, agent_type):
+    """Check RA/KRA Agent"""
+
+    try:
+        cert = x509.load_certificate_from_file(paths.RA_AGENT_PEM)
+    except Exception as e:
+        yield Result(plugin, constants.ERROR,
+                     error=str(e),
+                     msg='Unable to load RA cert: {error}')
+        return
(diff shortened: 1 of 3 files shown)


rcritten:
I think a better approach would to pull much of this out into a function that takes the DN to tell which entry to compare and return so a Result can be constructed. The is_installed check for the CA and RA/KRA will probably have to remain in the check itself for simplicity.

This will reduce a lot of code duplication.

A test would be nice too.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -724,6 +724,81 @@ def check(self):
                         self, constants.SUCCESS, key=cert)
 
 
+def check_agent(plugin, base_dn):
```
rcritten, line 727: If it took a string you could differentiate agent type:

e.g check_agent(plugin, base_dn, type):
...
logger.debug('%s agent descripiton should be %s', type, description)
...
        yield Result(plugin, constants.ERROR,
                     description=description,
                     msg='%s agent not found in LDAP' % type)

I guess it's implicit so this isn't a hard ask but it does read better IMHO, particularly in the human output.

rcritten:
Thinking about this further, I really think that the message should include KRA or RA for readability purposes.

antoniotorresm:
Yes, I added it as you suggested :)

rcritten:
Huh, I must have not reloaded the patch or something. In that case ack.

PR #193, opened 2021-04-23
Add checks to detect mismatch of certificates between databases

antoniotorresm:
Add checks to detect mismatch of certificates between LDAP
and NSS databases. Check for existance of entries as well as
ensure the certificates match between the different databases.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1886770

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -29,6 +29,7 @@ from ipaserver.plugins import ldap2
 from ipapython import certdb
 from ipapython import ipautil
 from ipapython.dn import DN
+from ipapython.ipaldap import realm_to_serverid
 
 logger = logging.getLogger()
 DAY = 60 * 60 * 24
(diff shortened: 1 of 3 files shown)


src/ipahealthcheck/ipa/certs.py
```diff
@@ -583,6 +584,103 @@ def check(self):
...
+        except Exception as e:
+            yield Result(self, constants.ERROR,
+                         error=str(e),
+                         msg='Unable to load CA cert: {error}')
+            return
+
+        # Ensure CA cert entry from LDAP matches /etc/ipa/ca.crt
+        try:
+            entry = self.conn.get_entry(DN('cn=CAcert,cn=ipa,cn=etc',
+                                           api.env.basedn))
```
rcritten, line 662: This should still be valid but it is the legacy location for storing the CA certificate. It lacks the full chain so may not be all that useful by itself.

The full chain is stored in cn=certificates,cn=ipa,cn=etc,$SUFFIX.

I suspect that with an external IPA install the check against ca.crt will fail.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -583,6 +584,103 @@ def check(self):
...
+                                  % (dbpath, paths.IPA_CA_CRT)))
+                return False
+            return True
+
+        db = certs.CertDB(api.env.realm, paths.PKI_TOMCAT_ALIAS_DIR)
+        dn = DN('uid=pkidbuser,ou=people,o=ipaca')
+        nick = 'subsystemCert cert-pki-ca'
+        subsystem_ok = yield from match_ldap_nss_cert(self, self.conn,
+                                                      db, dn,
+                                                      'userCertificate', nick)
```
rcritten, line 642: Arguably this check could be in src/ipahealthcheck/dogtag/ca.py since it's a dogtag cert.

Be sure to test this in a ca-less environment as well.

antoniotorresm:
I am not sure moving the Dogtag cert checks to `ca.py` is a good idea since LDAP access is needed, the plugin needs to inherit from IPAPlugin, not DogtagPlugin. I've separated it to a different class.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -583,6 +584,125 @@ def check(self):
...
+                return False
+            try:
+                nsscert = db.get_cert_from_db(cert_nick)
+            except Exception as e:
+                yield Result(plugin, constants.ERROR,
+                             error=str(e),
+                             msg=('Unable to load %s certificate:'
+                                  '{error}' % cert_nick))
+                return False
+            ldapcert = entry[attr][0]
```
rcritten, line 681: It's possible that there could be older, expired/revoked certificates in userCertificate. It would be safer to walk all the certs looking for a match.

It might also be more robust to ensure they are in the same format. For example load them as x509 objects and compare them rather than comparing the raw blobs.

src/ipahealthcheck/ipa/certs.py
```diff
@@ -583,6 +584,125 @@ def check(self):
...
+                                                      db, dn,
+                                                      'userCertificate', nick)
+        dn = DN('cn=%s IPA CA' % api.env.realm,
+                'cn=certificates,cn=ipa,cn=etc',
+                api.env.basedn)
+        nick = 'caSigningCert cert-pki-ca'
+        casigning_ok = yield from match_ldap_nss_cert(self, self.conn,
+                                                      db, dn, 'CACertificate',
+                                                      nick)
+        if subsystem_ok and casigning_ok:
```
rcritten, line 702: The Audit, OCSP and subystem certificates should also be checked. And if there is a KRA then those as well.

antoniotorresm:
Pushed again with proposed changes. Now all of CA/KRA entries are checked.

antoniotorresm:
The matching checks in CADogtagCertsConfigCheck are against CS.cfg, while the checks in this PR are against their LDAP entries. I don't think it is a duplicate check, both provide useful information. What do you think?

rcritten:
You're right, I misread it.

I opened https://github.com/dogtagpki/pki/issues/3518 against pki for the exceptions.

Can you add the nickname and dbdir as keywords to the msg and use variable substitution instead?
e.g.
```
    "kw": {
      "key": "ocspSigningCert cert-pki-ca",
      "nickname": "ocspSigningCert cert-pki-ca",
      "dbdir": "/etc/pki/pki-tomcat/alias",
      "msg": "{nickname} certificate in NSS DB {dbdir} does not match the entry in LDAP"
    }
```

Otherwise this is looking good.

antoniotorresm:
Done!

rcritten:
LGTM once these minor message issues are resolved. I tested with a CA-ful and CA-less install.

Note that testing with CA-less uncovered a couple of unrelated issues we'll need to address prior to next release.

antoniotorresm:
Message changes are already pushed :)

antoniotorresm:
Changes pushed.

antoniotorresm:
Pushed the change for looping through all certs in the LDAP entry for /etc/ipa/ca.crt.

antoniotorresm:
Pushed more changes proposed online. Now multiple SUCESS are returned too.

rcritten:
Nice catch on multiple SUCCESS.

I found just a few minor issues. This should correct them:

```
diff --git a/src/ipahealthcheck/ipa/certs.py b/src/ipahealthcheck/ipa/certs.p
y
index 767a217..afb731a 100644
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -598,7 +598,7 @@ class IPACertMatchCheck(IPAPlugin):
         """
         try:
             args = ["-L", "-n", nickname, "-a"]
-            result = nssdb.run_certutil(nssdb, args, capture_output=True)
+            result = nssdb.run_certutil(args, capture_output=True)
             return x509.load_certificate_list(result.raw_output)
         except ipautil.CalledProcessError:
             return []
@@ -799,7 +799,7 @@ class IPADogtagCertsMatchCheck(IPAPlugin):
                          key=casigning_nick)
         if ipaca_certs_ok:
             yield Result(self, constants.SUCCESS,
-                         key=ipaca_basedn)
+                         key=str(ipaca_basedn))
 
 
 @registry
diff --git a/tests/test_ipa_cert_match.py b/tests/test_ipa_cert_match.py
index ee5c897..460e61a 100644
--- a/tests/test_ipa_cert_match.py
+++ b/tests/test_ipa_cert_match.py
@@ -68,7 +68,7 @@ class mock_CertDB:
             raise errors.NotFound(reason='test')
         return IPACertificate()
 
-    def run_certutil(self, db, args, capture_output):
+    def run_certutil(self, args, capture_output):
         class RunResult:
             def __init__(self, output):
                 self.raw_output = output
```

antoniotorresm:
OK, pushed.

rcritten:
It still needs to fix for `key=str(ipaca_basedn))` on line ~799.

Run` ipa-healthcheck --source ipahealthcheck.ipa.certs --all `to see it

antoniotorresm:
OK, pushed.

rcritten:
Looks great, thanks for your patience.

ack.

PR #199, opened 2021-05-05
Add dependencies so healthcheck returns sooner if 389-ds is down

rcritten:
If dirsrv is down then not much works. Add a dependency to the IPAHostKeytab check and add dependency handling to the service checks.


fcami:
LGTM

fcami (approved):
ACK

PR #200, opened 2021-05-05
Restore the log level after loading the resources

rcritten:
It seems that loading the resources can affect the log level so
save it off before loading them and then restore it.

I'm not entirely sure what is happening but this is an easy fix
and we can address it further later if we want.


antoniotorresm:
LGTM, we should open a ticket with this so we make sure it is addressed eventually.

rcritten:
https://github.com/freeipa/freeipa-healthcheck/issues/210Good idea, filed

PR #202, opened 2021-05-11
Don't report CA errors in an IPA CA-less installation

rcritten:
The role checker was reporting CRL manager status. This isn't available in a CA-less install.

The pki healthcheck checks were all returning the same missing pki-tomcat instance error.

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -265,6 +265,13 @@ class RunChecks:
         if rval is not None:
             return rval
 
+        # If we have IPA configured without a CA then we want to skip
+        # the pkihealthcheck plugins otherwise they will generated a
+        # lot of false positives. The IPA plugins are loaded first so
+        # which should set ca_configured in its registry to True or
+        # False. We will skip the pkihealthcheck plugins only if
+        # ca_configured is False which means that it was set by IPA.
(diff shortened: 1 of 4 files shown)


src/ipahealthcheck/core/core.py
```diff
@@ -276,6 +277,11 @@ def run_healthcheck(self):
                 except Exception as e:
                     logger.error("Unable to initialize %s: %s", name, e)
                     continue
+            if hasattr(registry, 'ca_configured'):
+                ca_configured = registry.ca_configured
+            if 'pkihealthcheck' in name and ca_configured is False:
```
antoniotorresm, line 282: I think using `not ca_configured` instead of `ca_configured is False` would be better.

rcritten: The reason for this is that both None and False are True for not ca_configured. I suppose I can default to assuming that ca_configured is True instead of None.

rcritten: No, since this is in core this would likely break pkihealthcheck altogether as it would never be loaded. I think the original code is correct, I'll add some comments to clarify it.

antoniotorresm: OK, understood :)

Issue #203, opened 2021-05-13
Unhelpful error raised by IPACertDNSSAN with missing tracking cert


rcritten:
If an expected tracking cert is not present in the IPACertDNSSAN it throws a bad error. For one thing, we have no request_id because the request is not found. This also double-reports a missing cert.

I think it should debug log that the request is missing and move on.

Also, there is line-split issue which mashes tracked+by.

```
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertDNSSAN",
    "result": "ERROR",
    "uuid": "3aaa676b-d47b-4991-b45b-34d2bb612b0c",
    "when": "20210513124334Z",
    "duration": "9.922360",
    "kw": {
      "key": null,
      "msg": "Found request id {key} but it is not trackedby certmonger!?"
    }
  },
```

PR #204, opened 2021-05-13
Log and skip missing expected tracking in IPACertDNSSAN

rcritten:
This state is already reported by IPACertTracking so
avoid double-reporting it.

Also fix a missing space in a split line which caused
tracked by to display as trackedby.

https://github.com/freeipa/freeipa-healthcheck/issues/203


antoniotorresm:
LGTM.

rcritten:
Thanks for the review.

Issue #189, opened 2021-03-12
RFE: Add `--version` command line option


stanislavlevin:
It would be great to have the version indicator of the core as option to CLI.

PR #205, opened 2021-05-13
Add a --version cli option

rcritten:
I chose the pkg_resources method since it seemed the most
straightforward. It also gave me an excuse to configure
install_requires in setup.py. This works for me in a
development environment in Fedora anyway.

https://github.com/freeipa/freeipa-healthcheck/issues/189

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -157,6 +157,8 @@ def add_default_options(parser, output_registry, default_output):
                         default=default_output, help='Output method')
     parser.add_argument('--output-file', dest='outfile', default=None,
                         help='File to store output')
+    parser.add_argument('--version', dest='version', action='store_true',
+                        help='Report the version number and exit')
 
 
 def add_output_options(parser, output_registry):
@@ -242,6 +244,16 @@ class RunChecks:
         self.add_options()
         options = parse_options(self.parser)
         self.options = options
+
+        if options.version:
+            for registry in self.entry_points:
+                name = registry.split('.')[0]
+                try:
+                    version = pkg_resources.get_distribution(name).version
+                except pkg_resources.DistributionNotFound:
+                    continue
+                print('%s: %s' % (name, version))
+            return 0
 
         # pylint: disable=assignment-from-none
         rval = self.validate_options()
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,0 +1,15 @@
+#
+# Copyright (C) 2021 FreeIPA Contributors see COPYING for license
+#
+
+import os
+
+from ipapython.ipautil import run
+
+
+def test_version():
+    """
+    Test the --version option
+    """
+    output = run(['ipa-healthcheck', '--version'], env=os.environ)
+    assert 'ipahealthcheck' in output.raw_output.decode('utf-8')


stanislavlevin:
how about setuptools_scm? or for example, keep version in separate module and reference it in setup.py via `attr:`.

stanislavlevin:
pkg_resources is overkill there for self-resolve IMO.

rcritten:
pkg_resources is already invoked to resolve the plugins so we already have the overhead. Using it to pull the version before exiting is meh.

setuptools_scm is apparently not packaged for Fedora (so probably not RHEL either).

stanislavlevin:
just FYI https://src.fedoraproject.org/rpms/python-setuptools_scm

rcritten:
> just FYI https://src.fedoraproject.org/rpms/python-setuptools_scm

So odd. I did a dnf search setuptools_scm and got nothing but the package is clearly there.

stanislavlevin:
I think it would be usefull to print the versions of all installed plugins in the responce of `--version`, isn't it?

rcritten:
The 389-ds checks are a bit different since there is a wrapper around them so there is no easy-to-get version number.

The best I can do I think is:

ipahealthcheck: 0.8
pkihealthcheck: 0.1

src/ipahealthcheck/core/core.py
```diff
@@ -157,6 +157,9 @@ def add_default_options(parser, output_registry, default_output):
                         default=default_output, help='Output method')
     parser.add_argument('--output-file', dest='outfile', default=None,
                         help='File to store output')
+    parser.add_argument('--version', dest='version', action='store_true',
+                        default=False,
```
stanislavlevin, line 161: `default` is already False for `action='store_true`:
https://docs.python.org/3/library/argparse.html#action

stanislavlevin: `dest='version'` is also redundant.

stanislavlevin:
Ok, what about test?

src/ipahealthcheck/core/core.py
rcritten: Yeah I tend to just keep things explicit. I can remove it.

rcritten:
> Ok, what about test?

I added something, how great it is is up to interpretation. healthcheck does do a full multihost install like upstream IPA does so most of the integration testing is done there.

src/ipahealthcheck/core/core.py
stanislavlevin: Yes, please, remove.

src/ipahealthcheck/core/core.py
```diff
@@ -243,6 +246,15 @@ def run_healthcheck(self):
         options = parse_options(self.parser)
         self.options = options
 
+        if options.version:
+            for registry in self.entry_points:
+                name = registry.split('.')[0]
+                print(
+                    '%s: %s' %
+                    (name, pkg_resources.get_distribution(name).version)
```
stanislavlevin, line 254: This makes unconditional dependency on `pkihealthcheck` and fails if it's not installed:
```
E           ipapython.ipautil.CalledProcessError: CalledProcessError(Command ['ipa-healthcheck', '--version'] returned non-zero exit status 1: 'Traceback (most recent call last):\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/bin/ipa-healthcheck", line 8, in <module>\n    sys.exit(main())\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/ipahealthcheck/core/main.py", line 46, in main\n    sys.exit(ipachecks.run_healthcheck())\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/ipahealthcheck/core/core.py", line 254, in run_healthcheck\n    (name, pkg_resources.get_distribution(name).version)\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/pkg_resources/__init__.py", line 466, in get_distribution\n    dist = get_provider(dist)\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/pkg_resources/__init__.py", line 342, in get_provider\n    return working_set.find(moduleOrReq) or require(str(moduleOrReq))[0]\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/pkg_resources/__init__.py", line 886, in require\n    needed = self.resolve(parse_requirements(requirements))\n  File "/usr/src/RPM/BUILD/freeipa-healthcheck/.tox/py3/lib/python3/site-packages/pkg_resources/__init__.py", line 772, in resolve\n    raise DistributionNotFound(req, requirers)\npkg_resources.DistributionNotFound: The \'pkihealthcheck\' distribution was not found and is required by the application\n')
```

stanislavlevin: sorry for formatting.

tests/test_commands.py
```diff
@@ -0,0 +1,15 @@
...
+import os
+
+from ipapython.ipautil import run
+
+
+def test_version():
+    """
+    Test the --version option
+    """
+    output = run(['ipa-healthcheck', '--version'], env=os.environ)
```
stanislavlevin, line 14: no need to pass unmodified `env`, will be copied from current by default.

rcritten: ipautil.run clears PATH if environment isn't passed in which breaks a virtual environment.

src/ipahealthcheck/core/core.py
rcritten: I'm not sure when this may happen but I catch and ignore the exception now.

tests/test_commands.py
stanislavlevin: Not obvious behavior for me, but OK.

src/ipahealthcheck/core/core.py
stanislavlevin: This happens when I run freeipa-healthcheck self-tests and there is no dependency on pki-healthcheck right now.

stanislavlevin: `core` package shouldn't have deps on plugins.

stanislavlevin:
Awesome, works as expected:
- core only
```console
(py3) [builder@localhost freeipa-healthcheck]$ ipa-healthcheck --version
ipahealthcheck: 0.8
```

- pki plugin
```console
(py3) [builder@localhost freeipa-healthcheck]$ ipa-healthcheck --version
ipahealthcheck: 0.8
pkihealthcheck: 0.1
```

May be in future 389 team will add the version for plugin.

LGTM, thank you!

rcritten:
Thanks for the review.

PR #206, opened 2021-05-13
Replace the soon-to-be EOL F32 container with F34

rcritten:

### DIFF
--- a/.github/workflows/pipelines.yml
+++ b/.github/workflows/pipelines.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


stanislavlevin:
@rcritten, hi!
Thank you for PR!
All is Ok, just a note. F32 has Python3.8, while F33 and F34 have Python3.9+,
should Python3.8  be dropped?
```diff
diff --git a/.github/workflows/pipelines.yml b/.github/workflows/pipelines.yml
index 9c82d48..613e95a 100644
--- a/.github/workflows/pipelines.yml
+++ b/.github/workflows/pipelines.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9]
+        python-version: [3.9]
 
     steps:
     - uses: actions/checkout@v2

```

rcritten:
Good point, dropping 3.8

I probably need to re-test to figure out what the minimum version of python is supported. It is still set to 3.6, but that is unrelated to this.

stanislavlevin:
LGTM! Thank you!

stanislavlevin:
If this project follows FreeIPA then it should be Python3.6.

rcritten:
It's tricky because while I don't want to do anything to actively break older pythons they aren't well-tested with either these unit tests nor with freeIPA PR-CI.

I'll leave it as-is for now.

Issue #207, opened 2021-05-14
3rd party CA certificates output non-json 


arnoxit:
`ipahealthcheck.dogtag.ca` and `ipahealthcheck.ipa.certs` output text warning that additional CA certificates are not found and that it is assuming they are 3rd party, which in our case is correct, but these messages come out as plain strings to stdout and so break json validation.
```
# ipa-healthcheck --failures-only
ISRGRootX1 not found, assuming 3rd party
LetsencryptRootR3 not found, assuming 3rd party
LetsencryptRootR4 not found, assuming 3rd party
UserTrustRSARoot not found, assuming 3rd party
ZeroSSLRSADomainSecureSiteCA not found, assuming 3rd party
BPClass2Rot not found, assuming 3rd party
ISRGRootX2 not found, assuming 3rd party
LetsencryptRootE2 not found, assuming 3rd party
DSTRootX3 not found, assuming 3rd party
BPClass2CA5 not found, assuming 3rd party
LetsencryptRootE1 not found, assuming 3rd party
ISRGRootX1 not found, assuming 3rd party
LetsencryptRootR3 not found, assuming 3rd party
LetsencryptRootR4 not found, assuming 3rd party
UserTrustRSARoot not found, assuming 3rd party
ZeroSSLRSADomainSecureSiteCA not found, assuming 3rd party
BPClass2Rot not found, assuming 3rd party
ISRGRootX2 not found, assuming 3rd party
LetsencryptRootE2 not found, assuming 3rd party
DSTRootX3 not found, assuming 3rd party
BPClass2CA5 not found, assuming 3rd party
LetsencryptRootE1 not found, assuming 3rd party
[
  {
    "source": "ipahealthcheck.ipa.dna", 
    "kw": {
      "msg": "No DNA range defined. If no masters define a range then users and groups cannot be created.", 
      "range_start": 0, 
      "next_start": 0, 
      "next_max": 0, 
      "range_max": 0
    }, 
    "uuid": "60e2355a-8426-433b-84fe-f4d42bd92be6", 
    "duration": "0.371380", 
    "when": "20210514151132Z", 
    "check": "IPADNARangeCheck", 
    "result": "WARNING"
  }
]
```

I would expect these to be json formatted WARNINGs (as there is no INFO) or at least sent to stderr so they don't get mixed with the json output.

arnoxit:
[...]

PR #208, opened 2021-05-17
Replace or drop print() statements

rcritten:
The print in dogtag/ca.py was supposed to be a debug statement.
Converted it to use logger and fixed a related issue that was
calling logging.debug instead of logger.debug.

Drop a print() statement from a test, probably leftover from
troubleshooting a failure.

https://github.com/freeipa/freeipa-healthcheck/issues/207


antoniotorresm:
There is another `logging.debug` in `ipahealthcheck/ipa/plugin.py`.

rcritten:
Thanks, fixed.

antoniotorresm:
LGTM.

antoniotorresm:
LGTM.

Issue #213, opened 2021-06-14
File checker doesn't handle non-existent files


rcritten:
Some files that are important to IPA may not always exist, particularly log files.

[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "check": "IPAFileCheck",
-- | -- | --
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "result": "CRITICAL",
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "uuid": "ef235900-2a7c-482c-aa13-7f27c6e9c3ce",
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "when": "20210613034705Z",
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "duration": "0.048699",
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "kw": {
[08:11] | fcami | DEBUG    ipatests.pytest_ipa.integration.host.Host.master.cmd65:transport.py:557 "exception": "[Errno 2] No such file or directory: '/var/log/pki/pki-tomcat/kra/selftests.log'"

PR #214, opened 2021-06-14
Handle files that don't exist in FileCheck

rcritten:
A raw os.stat() was called which could raise an exception if the file
doesn't exist. Instead call os.path.exists() and if the result is False
then raise a SUCCESS with a message that the file doesn't exist.

https://github.com/freeipa/freeipa-healthcheck/issues/213


fcami (approved):
Thanks for the quick turnaround.
ACK

Issue #215, opened 2021-06-18
[pylint] new recommendations


stanislavlevin:
Minor recommendations from `pylint==2.8.2, astroid==2.5.6`
```
************* Module src.ipahealthcheck.ipa.certs
src/ipahealthcheck/ipa/certs.py:722:27: R1729: Use a generator instead 'any(cert == nsscert for cert in entry[attr])' (use-a-generator)
src/ipahealthcheck/ipa/certs.py:739:21: R1729: Use a generator instead 'any(cert in entry['userCertificate'] and subject == entry['subjectName'][0] for entry in entries if 'userCertificate' in entry)' (use-a-generator)

------------------------------------
Your code has been rated at 10.00/10
```

stanislavlevin:
Running latest Pylint against `master` shows 'error':
```
************* Module tests.test_ipa_trust
tests/test_ipa_trust.py:37:0: E0611: No name 'OPT_X_SASL_SSF_MIN' in module 'ldap' (no-name-in-module)
************* Module tests.test_ipa_agent
tests/test_ipa_agent.py:18:0: E0611: No name 'OPT_X_SASL_SSF_MIN' in module 'ldap' (no-name-in-module)
```

PR #216, opened 2021-06-18
pylint: Fix new recommendations

stanislavlevin:
- use-a-generator (R1729)
> Use a generator instead '%s(%s)' Comprehension inside of 'any' or
'all' is unnecessary. A generator would be sufficient and faster.

- http://pylint.pycqa.org/en/latest/technical_reference/c_extensions.html

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/215

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,6 +1,9 @@
 [MASTER]
 persistent=no
 
+extension-pkg-whitelist=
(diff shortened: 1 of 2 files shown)


rcritten:
LGTM, thanks for the patch.

Issue #217, opened 2021-06-28
IPATrustControllerServiceCheck doesn't handle HIDDEN_SERVICE


rcritten:
The source ipahealthcheck.ipa.trust check IPATrustControllerServiceCheck checks whether the ADTRUST service is an ENABLED_SERVICE.

It also needs to consider HIDDEN_SERVICE to be an acceptable state.

rcritten:
RHBZ https://bugzilla.redhat.com/show_bug.cgi?id=1976878

PR #218, opened 2021-06-28
Allow for HIDDEN_SERVICE when checking ADTRUST service

rcritten:
If the host is a trust controller then the ADTRUST service
must be enabled. This is defined as both ENABLED_SERVICE and
HIDDEN_SERVICE.

https://github.com/freeipa/freeipa-healthcheck/issues/217


fcami (approved):
LGTM. ACK even.

PR #223, opened 2021-09-03
Fix some deprecation issues, add support for DNS URI records

rcritten:
Fix a couple of deprecation warnings coming from IPA and python3-dns.

Also add support for DNS URI records. If the underlying IPA during the test execution contains support for URI records but that support isn't built into healthcheck then some tests will fail and output like this will be seen:

ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256
ERROR    root:idns.py:73 Unhandled rdtype 256

rdtype 256 == URI

So do this as an all-in-one shot since the DNS changes are related.

### DIFF
--- a/pylint_plugins.py
+++ b/pylint_plugins.py
@@ -1,30 +1,510 @@
-""" Plugin to teach Pylint about FreeIPA API
-"""
+#
+# Copyright (C) 2015  FreeIPA Contributors see COPYING for license
+#
 
+from __future__ import print_function
+
+import copy
+import os.path
+import sys
 import textwrap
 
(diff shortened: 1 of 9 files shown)


fcami (approved):
ACK

Issue #224, opened 2021-09-03
Include tracebacks in the Result when catching outer-level problems


rcritten:
Healthcheck has a fail-safe so that if any of the checks raise an exception it is caught and logged so that one can be sure that all checks have run, even if not successfully. The problem is that the traceback is suppressed and often the exception message isn't very useful. Log the traceback both in debug mode and in the result itself to make tracking down the root cause a lot easier.

PR #225, opened 2021-09-03
Include an exception on outer-level failures when running plugins

rcritten:
The traceback will be both in the debug output and within the
Result value for the call so we have half a chance to fix what
is broken.

https://github.com/freeipa/freeipa-healthcheck/issues/224


rcritten:
This requires the pylint changes contained in https://github.com/freeipa/freeipa-healthcheck/pull/223

fcami (commented):
hi @rcritten could you rebase and re-push please

Issue #222, opened 2021-09-03
Add support for the DNS URI type in IPA


rcritten:
IPA included support for the DNS URI system records for the KDC In f1c58fb6466fba516f3d3e50a1a982b0162ef902

healthcheck should verify these if present. Skip them otherwise for backwards compatibility,.

rcritten:
Merged.

PR #226, opened 2021-09-17
URI records are not required but if they exist they are validated.

rcritten:
This fixes support when testing with a version of IPA that does not
include URI support.

https://github.com/freeipa/freeipa-healthcheck/issues/222


rcritten:
I'm not sure the motivation for this patch as it duplicates code that is already there. Closing.

PR #227, opened 2021-09-17
Unify command-line options and configuration

rcritten:
This adds command-line options to the configuration, but not vice-versa.
It will allow for greater flexibility when running this automatically
without having to change any scripting.

https://bugzilla.redhat.com/show_bug.cgi?id=1872467

### DIFF
--- a/man/man5/ipahealthcheck.conf.5
+++ b/man/man5/ipahealthcheck.conf.5
@@ -36,6 +36,14 @@ The following options are relevant for the server:
 .TP
 .B cert_expiration_days\fR
 The number of days left before a certificate expires to start displaying a warning. The default is 28.
+.TP
+All command\-line options may be included in the configuration file. Dashes must be converted to underscore for the configuration file, e.g. \-\-output\-type becomes output_type. All options, including those that don't make sense in a config file, like \-\-list\-sources, are allowed. Let the buyer beware.
+.TP
+The purpose of allowing command\-line options to be in the configuration file is for automation without having to tweak the automation script. For example, if you want the default output type to be human for the systemd timer automated runs, settting output_type=human in the configuration file will do this. When loading configuration the first option wins, so if any option is in the configuration file then it cannot be overridden by the command-line unless a different configuration file is specified (see \-\-config).
+.TP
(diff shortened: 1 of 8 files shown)


rcritten:
Added a unit test for verifying that the configuration file options take precedence.

src/ipahealthcheck/core/config.py
```diff
@@ -63,7 +63,7 @@ def merge(self, d):
         """
         Merge variables from dict ``d`` into the configuration
 
-        The last one wins.
+        The first one wins.
```
antoniotorresm, line 66: Isn't it the last one? If both dicts have the same key, the value will be updated to the value for the key in the passed dict.

rcritten: This is correct. The reason is because of the default cli options. I wasn't able to find a reliable way to tell whether the option had been passed in vs the argparse default value. Since pki also uses the framework I can't drop all default values because they may use them as well.

The tests verify this in an indirect way. e.g. setting output_type=human vs the default value of json.

antoniotorresm: But I don't understand the comment. Isn't the last one winning?

rcritten: It's because of the `object.__setattr__(self, key, value)` in __setitem__. It basically makes it immutable.

antoniotorresm: Thanks, missed that.

antoniotorresm (approved):
LGTM.

man/man5/ipahealthcheck.conf.5
```diff
@@ -36,6 +36,14 @@ The following options are relevant for the server:
 .TP
 .B cert_expiration_days\fR
 The number of days left before a certificate expires to start displaying a warning. The default is 28.
+.TP
+All command\-line options may be include in the configuration file. Dashes must be converted to underscore for the configuration file, e.g. \-\-output\-type becomes output_type. All options, including those that don't make sense in a config file, like \-\-list\-sources, are allowed. Let the buyer beware.
```
fcami, line 40: included

fcami (approved):
See above, otherwise LGTM.

man/man5/ipahealthcheck.conf.5
rcritten: fixed, thanks

Issue #176, opened 2021-01-04
RFE: Add --exclude parameter


cipherboy:
`ipa-healthcheck` executes a decent number (11) of check by default (i.e., no paramters) on my box:

<details>
    <summary>Example output</summary>

```[
  {
    "source": "ipahealthcheck.ds.backends",
    "check": "BackendsCheck",
    "result": "CRITICAL",
    "uuid": "126c6ec0-5207-4d02-8bc7-a77483028bb6",
    "when": "20210104191018Z",
    "duration": "0.020068",
    "kw": {
      "exception": "No object exists given the filter criteria cn=changelog"
    }
  },
  {
    "source": "ipahealthcheck.ipa.idns",
    "check": "IPADNSSystemRecordsCheck",
    "result": "WARNING",
    "uuid": "59973f5a-9804-4a7c-a7ae-c37f10aab9e8",
    "when": "20210104191027Z",
    "duration": "5.196823",
    "kw": {
      "msg": "Expected SRV record missing",
      "key": "_ldap._tcp.cipherdev.io.:ipa.cipherdev.io."
    }
  },
  {
    "source": "ipahealthcheck.ipa.idns",
    "check": "IPADNSSystemRecordsCheck",
    "result": "WARNING",
    "uuid": "6b9f458e-a9ce-4b9a-acda-5661402a9987",
    "when": "20210104191030Z",
    "duration": "8.550971",
    "kw": {
      "msg": "Expected SRV record missing",
      "key": "_kerberos._tcp.cipherdev.io.:ipa.cipherdev.io."
    }
  },
  {
    "source": "ipahealthcheck.ipa.idns",
    "check": "IPADNSSystemRecordsCheck",
    "result": "WARNING",
    "uuid": "e53645cd-87be-4e89-9a47-0ebd6e3df17c",
    "when": "20210104191031Z",
    "duration": "9.503003",
    "kw": {
      "msg": "Expected SRV record missing",
      "key": "_kerberos._udp.cipherdev.io.:ipa.cipherdev.io."
    }
  },
  {
    "source": "ipahealthcheck.ipa.idns",
    "check": "IPADNSSystemRecordsCheck",
    "result": "WARNING",
    "uuid": "ba07c16b-54cd-40c2-8f14-e43247d0edf5",
    "when": "20210104191034Z",
    "duration": "12.019969",
    "kw": {
      "msg": "Expected SRV record missing",
[...]

PR #228, opened 2021-09-22
Add a way to exclude sources, checks and/or keys from results

rcritten:
We allow to exclude on three different levels:

 * source
 * check
 * key

Excluding a source could be used to disable a misbehaving set of
checks, particularly one not provided by upstream. A check would
be similar.

Not all results have a key but most do. If we run into corner
cases we can address them as they come up.

The example I had in mind is an untracked certificate that is
otherwise legitimate. This could be marked as excluded by key
so ipa-healthcheck will no longer return failures.

Filtering happens twice. Any sources or checks excluded will simply
not be executed. keys are excluded after execution.

This adds new section, [exclusions], which will contain three types
of exclusions and can be repeated:

source
config
key

https://github.com/freeipa/freeipa-healthcheck/issues/176


rcritten:
An example config file may look like:

```
[default]

[excludes]
#source=ipahealthcheck.ipa.certs
#check=IPACertTracking
#check=IPADNSSystemRecordsCheck
key=20210910141452
```
So you can have source=ipahealthcheck.ipa to block out all but the pki healthchecks.

man/man5/ipahealthcheck.conf.5
```diff
@@ -31,11 +31,33 @@ Values should not be quoted, the quotes will not be stripped.
 Options must appear in the section named [default]. There are no other sections defined or used currently.
 
 Options may be defined that are not used. Be careful of misspellings, they will not be rejected.
+.SH "EXCLUDES"
+There may be reasons that a user will want to suppress some results. One example is a customer certificate that is generating a warning because it is unknown to IPA. Excluding a result byu key does not prevent it from running, it is filtered from the reported results. Excluding by source or key will prevent it from running at all. Services will not be excluded because other checks may rely on them (ipahealthcheck.meta.services).
```
antoniotorresm, line 35: Typo: "byu"

src/ipahealthcheck/core/config.py
```diff
@@ -5,13 +5,23 @@
...
-from ipahealthcheck.core.constants import CONFIG_SECTION
+from ipahealthcheck.core.constants import CONFIG_SECTION, EXCLUDE_SECTION
 from ipahealthcheck.core.constants import DEFAULT_CONFIG
 
 logger = logging.getLogger()
 
 
+class DuplicateOrderedDict(OrderedDict):
+    def __setitem__(self, key, value):
+        """Duplicate keys will be concatonated strings separated by new-line"""
```
antoniotorresm, line 18: Typo: "concatonated"

antoniotorresm:
Is it possible to use `excludes_key`, `excludes_source` and `excludes_check` options outside the `[excludes]` section? The man update gives that impression, but testing on my side fails:
```

[default]
excludes_key=DSVIRTLE0001

```
```
Traceback (most recent call last):
  File "/root/freeipa-healthcheck/venv/bin/ipa-healthcheck", line 33, in <module>
    sys.exit(load_entry_point('ipahealthcheck', 'console_scripts', 'ipa-healthcheck')())
  File "/root/freeipa-healthcheck/src/ipahealthcheck/core/main.py", line 52, in main
    sys.exit(ipachecks.run_healthcheck())
  File "/root/freeipa-healthcheck/src/ipahealthcheck/core/core.py", line 420, in run_healthcheck
    results = exclude_keys(config, results)
  File "/root/freeipa-healthcheck/src/ipahealthcheck/core/core.py", line 256, in exclude_keys
    result.kw.get("key") in config.excludes_key
TypeError: 'in <string>' requires string as left operand, not NoneType
```

antoniotorresm:
Do you think it would be useful to have these options on the CLI as well?

antoniotorresm:
Maybe you could add an example usage to the ipahealtcheck.conf man page. Other than that this LGTM.

rcritten:
Man page updated.

Issue #231, opened 2021-11-05
Compare proxy shared secret configuration


rcritten:
IPA uses a shared secret to configure the proxy to the CA. There are two possible issues:

1. The shared secret differs between the two. This will lead to a cryptic 403 error that is often confused with an RA agent mismatch.
2. In CentOS 8 there is an issue where both secret and requiredSecret are set in the tomcat server.xml configuration and the secrets don't match what is in ipa-pki-proxy.conf.

PR #234, opened 2021-11-08
Compare proxy shared secret configuration

rcritten:
Compare the ProxyPassMatch secret(s) with those in server.xml

For now we are skipping checking to see if both secret and
requiredSecret are configured since it doesn't seem to cause
tomcat any issues. As long as the secrets match up with
ipa-pki-proxy.conf then things work fine.

https://github.com/freeipa/freeipa-healthcheck/issues/231


antoniotorresm:
This looks good to me.

Issue #233, opened 2021-11-05
Ensure the apache user a member of the ipaapi group


rcritten:
This is part of the privilege separation and will allow Apache to read/write the stored ccaches.

Without being in this group Apache will not be able to read and store session information and return cryptic error messages to the user, either "all servers failed" or an error about missing Negotiate header.

PR #235, opened 2021-11-15
Check expected group memberships

rcritten:
The initial purpose of this is for privilege separation
where if the group membership is not correct the ccaches
may not be readable. It's possible this will expand to other
purposes.

https://github.com/freeipa/freeipa-healthcheck/issues/233

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,6 +44,7 @@ setup(
             'ipafiles = ipahealthcheck.ipa.files',
             'ipahost = ipahealthcheck.ipa.host',
             'ipameta = ipahealthcheck.ipa.meta',
+            'ipanss = ipahealthcheck.ipa.nss',
             'iparoles = ipahealthcheck.ipa.roles',
             'ipatopology = ipahealthcheck.ipa.topology',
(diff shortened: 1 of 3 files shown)


antoniotorresm:
This is failing for me on a freshly installed IPA 4.9.8 server on Fedora 35.

```
  {
    "source": "ipahealthcheck.ipa.nss",
    "check": "IPAGroupMemberCheck",
    "result": "ERROR",
    "uuid": "315f850e-b192-4053-a1d0-09f0a61c9f30",
    "when": "20220114131944Z",
    "duration": "0.000080",
    "kw": {
      "key": "apache",
      "member": "ipaapi",
      "msg": "{member} is not a member of group {key}"
    }
  }
]
```

`apache` user is part of the `ipaapi` group:
```
(venv) [root@ci-vm-10-0-138-110 freeipa-healthcheck]# groups apache
apache : apache ipaapi
```

rcritten:
I can only assume that during testing I flipped the members tuple. It was correct in the documentation block but incorrect in code. I flipped it back and things should work now.

antoniotorresm:
Yes, it's working fine now. LGTM.

Issue #237, opened 2021-11-30
[core] too strict dependency on ipaserver


stanislavlevin:
Currently `core` plugin system unconditionally requires `ipaserver`
`src/ipahealthcheck/core/main.py`:
```python
from ipaserver.install.installutils import is_ipa_configured
```

This means that all the FreeIPA stack should be installed to get the configuration status of IPA server even if ipaserver package is not installed.

I suggest making this import optional and treating the missing `ipaserver` as unconfigured installation.

rcritten:
I see. So a healthcheck that is used by something outside the IPA universe would still need IPA installed?

stanislavlevin:
Not practically yet, it's more about modularity and packaging. My point is that the core system ideally should not depend on ipa stuff, but the freeipa-healthcheck plugin itself. For example, being reusable the core may be called outside of console script (`ipa-healthcheck`) by any lib/application (if ipaserver is not installed):
```console
[root@281a5762c1bd /]# python3 -c 'import ipahealthcheck.core.main as main; main.main()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/ipahealthcheck/core/main.py", line 12, in <module>
    from ipaserver.install.installutils import is_ipa_configured
ModuleNotFoundError: No module named 'ipaserver'
```

Actual problem is that every plugin of healthcheck system gains IPA stack as an indirect dependency even if IPA is not used by a plugin. For example, in ALTLinux `dogtag-pki-server` requires `freeipa-healthcheck-core` which in turn, wants ipaserver, but actually dogtag-pki-server works just fine without the latter. Moreover, this can lead to build loops known as bootstrap issues like
dogtag -> healthcheck -> ipa -> dogtag

rcritten:
Yeah, makes sense. The "core" stuff was basically stuffed in after the initial design so its still sort of bolted on. Adding a try/except on the import is not a big deal and a good idea. I probably won't get to this until next calendar year though.

PR #238, opened 2021-12-01
core: Relax dependency on IPA stuff

stanislavlevin:
The core plugin system ideally should not depend on IPA, but the
freeipa-healthcheck plugin itself. For example, being reusable the
core may be called outside of console script (ipa-healthcheck) by
any lib/application (if ipaserver is not installed):

```console
[root@281a5762c1bd /]# python3 -c 'import ipahealthcheck.core.main as main; main.main()'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/lib/python3.10/site-packages/ipahealthcheck/core/main.py", line 12, in <module>
    from ipaserver.install.installutils import is_ipa_configured
ModuleNotFoundError: No module named 'ipaserver'
```

Actual problem is that every plugin of healthcheck system gains extra
IPA stack as the indirect dependency even if IPA is not used by a plugin.
For example, in ALTLinux dogtag-pki-server requires freeipa-healthcheck-core
which in turn, wants ipaserver, but actually dogtag-pki-server works
just fine without the latter. Moreover, this can lead to build loops
known as bootstrap issues like `dogtag -> healthcheck -> ipa -> dogtag`.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/237

### DIFF
--- a/src/ipahealthcheck/core/main.py
+++ b/src/ipahealthcheck/core/main.py
@@ -8,11 +8,18 @@ import sys
 from ipahealthcheck.core import constants
 from ipahealthcheck.core.core import RunChecks
 
-from ipalib.facts import is_ipa_configured
+try:
+    from ipalib.facts import is_ipa_configured
+except ImportError:
+    is_ipa_configured = None
 
 
 class IPAChecks(RunChecks):
     def pre_check(self):
+        if is_ipa_configured is None:
+            print("IPA server is not installed")
+            return 1
+
         if not is_ipa_configured():
             print("IPA server is not configured")
             return 1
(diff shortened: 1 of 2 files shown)


stanislavlevin:
For the record, the issue and the commit message references to `ipaserver`, actually it should be `ipalib`. This is due to the changes between release(0.9) and the current master. This fact doesn't change the rationalle.

rcritten:
Thanks for the patch, looks great.

stanislavlevin:
@rcritten, thanks for the review!

Issue #236, opened 2021-11-24
Ability to lower the timeout for some replication checks


baryluk:
Hi,

we are considering using https://github.com/camptocamp/ipahealthcheck_exporter for monitoring of some of our IPA servers. It uses `ipa-healthcheck` internally, but I find it extremely slow and not well suited for the real time monitoring task.

One of the reasons is that `ipa-healtcheck` is non-local.

We would like to run it in a mode, where it does not contact any other servers, and only reports status based on locally available information (i.e. ipa server already knows its own state of the replication, and state of its peers, so ipa-healtcheck does not need to recheck it).

Second but related issue, is that `ipa-healthcheck` contacts other replicas, but if they are down, it takes about 2 minutes for the checks to timeout, which makes the entire check to take about 8 minutes to finish.

We would like to either disable these remote checks, or reduce timeout to about 10 seconds.

Example of checks that are very slow:

```json
[
...
  {
    "source": "ipahealthcheck.ds.replication",
    "check": "ReplicationCheck",
    "result": "ERROR",
    "uuid": "2b03e49b-7c70-45f7-9c40-3bf3cb7afd8a",
    "when": "20211124100425Z",
    "duration": "261.222873",
    "kw": {
      "key": "DSREPLLE0005",
      "items": [
        "Replication",
        "Agreement"
      ],
      "msg": "The replication agreement (foo.example.com-to-bar.example.com) under \"dc=example,dc=com\" is not in synchronization,\nbecause the consumer server is not reachable."
    }
  },

...
  {
    "source": "pki.server.healthcheck.clones.connectivity_and_data",
    "check": "ClonesConnectivyAndDataCheck",
    "result": "ERROR",
    "uuid": "37858400-a560-4e91-815c-1cc5ed6b0c1c",
    "when": "20211124100214Z",
    "duration": "131.126375",
    "kw": {
      "status": "ERROR:  pki-tomcat : Internal error testing CA clone. Host: bar.example.com Port: 443"
    }
  },
  {
    "source": "pki.server.healthcheck.clones.connectivity_and_data",
    "check": "ClonesConnectivyAndDataCheck",
    "result": "SUCCESS",
    "uuid": "f4d7f33b-76fe-4dee-872d-303753a40c60",
    "when": "20211124100214Z",
    "duration": "131.126410",
    "kw": {
      "instance_name": "pki-tomcat",
      "status": "KRA Clones tested successfully, or not present."
    }
  },
  {
[...]

PR #239, opened 2021-12-03
Add a timeout wrapper around each check execution

rcritten:
A timeout will raise a new exception, TimeoutError. This
can be caught and handled inside an individual check, otherwise
it will be handled by run_plugin.

https://github.com/freeipa/freeipa-healthcheck/issues/236


antoniotorresm (approved):
This looks good to me.

rcritten:
Patch rebased, conflict in ipahealthcheck.conf man page.

PR #240, opened 2021-12-08
add support for prometheus text metric exposition format output.

UiP9AV6Y:
this new output plugin generates metrics from the check results in the prometheus
text exposition format. it is intended to be used in combination with the
prometheus node_exporter [textfile collector][]. the output plugin generates
similar metrics as the [ipahealthcheck_exporter][].

[textfile collector]: https://github.com/prometheus/node_exporter#textfile-collector
[ipahealthcheck_exporter]: https://github.com/camptocamp/ipahealthcheck_exporter

an example output of the plugin:

```
# HELP ipa_cert_expiration Expiration date of certificates in warning/error state
# TYPE ipa_cert_expiration gauge
ipa_cert_expiration{certificate_request_id="20200416134554"} 1.618580754e+09
# HELP ipa_service_state State of the services monitored by IPA healthcheck
# TYPE ipa_service_state gauge
ipa_service_state{service="certmonger"} 1.0
ipa_service_state{service="dirsrv"} 1.0
ipa_service_state{service="gssproxy"} 1.0
ipa_service_state{service="httpd"} 1.0
ipa_service_state{service="ipa_custodia"} 1.0
ipa_service_state{service="ipa_dnskeysyncd"} 1.0
ipa_service_state{service="ipa_otpd"} 1.0
ipa_service_state{service="kadmin"} 1.0
ipa_service_state{service="krb5kdc"} 1.0
ipa_service_state{service="named"} 1.0
ipa_service_state{service="pki_tomcatd"} 1.0
ipa_service_state{service="sssd"} 1.0
# HELP ipa_healthcheck Number of healthchecks with a certain result
# TYPE ipa_healthcheck gauge
ipa_healthcheck{result="SUCCESS"} 244.0
ipa_healthcheck{result="WARNING"} 1.0
```

### DIFF
--- a/man/man8/ipa-healthcheck.8
+++ b/man/man8/ipa-healthcheck.8
@@ -37,7 +37,7 @@ Execute checks within the named source, or all sources in the given namespace.
 Execute this particular check within a source. The exact source must also be specified via \fB\-\-source\fR.
 .TP
 \fB\-\-output\-type\fR=\fITYPE\fR
-Set the output type. Supported variants are \fBhuman\fR and \fBjson\fR. The default is \fBjson\fR.
+Set the output type. Supported variants are \fBhuman\fR, \fBjson\fR, and \fBprometheus\fR. The default is \fBjson\fR.
 .TP
 \fB\-\-failures\-only\fR
 Exclude SUCCESS results on output. If stdin is a tty then this will default to True. In all other cases it defaults to False.
@@ -71,6 +71,14 @@ The results are displayed in a more human\-readable format.
 .TP
 \fB\-\-input\-file\fR=\fIFILENAME\fR
 Take as input a JSON results output and convert it to a more human\-readable form.
+
+.SS "PROMETHEUS OUTPUT"
+The results are displayed in the Prometheus text metric exposition format.
+.TP
+\fB\-\-input\-file\fR=\fIFILENAME\fR
+Uses the JSON-formatted results output as metrics source.
+\fB\-\-metric\-prefix\fR=\fIPREFIX\fR
+Prefix to use for metric names.
 
 .SH "EXAMPLES"
 .PP
--- a/src/ipahealthcheck/core/output.py
+++ b/src/ipahealthcheck/core/output.py
@@ -2,6 +2,7 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 
+from datetime import datetime
 import json
 import sys
 from ipahealthcheck.core.constants import _nameToLevel, SUCCESS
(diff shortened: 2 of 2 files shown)


rcritten:
Thanks for the PR!

A couple of maintenance nits. Can you add a signed-off-by to the commit and expand on the commit message itself? I suppose that what is included as the description is sufficient.

How often is something like this executed? ipa-healthcheck is pretty lightweight but can take a few seconds to run all the checks and it is likely to get slower over time. If a process is pinging it hard it is likely to fail.

You are filtering after the run, which I suppose is the only way to do it currently. This means it is running a bunch of checks that won't be reported on. It never occurred to me to have an Output plugin manage the checks run but I suppose for a monitoring system it makes sense. As long as you are ok with the limitation for now I have no major objections.

rcritten:
The linting fails with:

 src/ipahealthcheck/core/output.py:169: [E0401(import-error), Prometheus.generate] Unable to import 'prometheus_client')

I wonder if this plugin can be made optional if the import is available.

UiP9AV6Y:
> The linting fails with:
> 
> src/ipahealthcheck/core/output.py:169: [E0401(import-error), Prometheus.generate] Unable to import 'prometheus_client')
> 
> I wonder if this plugin can be made optional if the import is available.

would you prefer something along the line of a conditional class declaration?

```python
try:
    from prometheus_client import CollectorRegistry, Gauge, generate_latest

    has_prometheus_client = True
except ImportError:
    has_prometheus_client = False

if has_prometheus_client:
    @output_registry
    class Prometheus(Output):
    ....
```

UiP9AV6Y:
> A couple of maintenance nits. Can you add a signed-off-by to the commit and expand on the commit message itself? I suppose that what is included as the description is sufficient.

fixed

> How often is something like this executed? ipa-healthcheck is pretty lightweight but can take a few seconds to run all the checks and it is likely to get slower over time. If a process is pinging it hard it is likely to fail.

we plan to run this every 5 minutes. IMO apart from the service checks, nothing
it real-time critical, i.e. it should not matter if an alert is generated now or in 5 minutes

> You are filtering after the run, which I suppose is the only way to do it currently. This means it is running a bunch of checks that won't be reported on. It never occurred to me to have an Output plugin manage the checks run but I suppose for a monitoring system it makes sense. As long as you are ok with the limitation for now I have no major objections.

i inspect checks as some of them generate additional metrics. this is mostly done
to be compatible with [ipahealthcheck_exporter](https://github.com/camptocamp/ipahealthcheck_exporter). i do *not* ignore check results itself, only their message and additional meta data. the *ipa_healthcheck* metric is a sum of all checks with a certain result. i was looking into a more generic processing solution (e.g.
automatically generate metrics from numeric metadata), but the check data has
little to no semantic, which makes metric generation kind of complicated (e.g.
defining the type of prometheus metric: gauge, counter, ...) and therefor would
require manual intervention anyway.
but if someone wants to generate more metrics, the plugin can certainly be extended
to do so.

UiP9AV6Y:
i have refactored the output plugin to work without any third-party dependencies. mainly so package maintainers dont end up with surprises when the new output plugin does not work despite the metadata (setup.py, requirements.txt, ..) not mentioning prometheus_client as dependency (i do not expect package maintainers to search the code for potential dependencies; at least i do not)

rcritten:
This looks good to me. In the future if you add more output you may want to consider a helper, or a template, to generate the HELP and TYPE lines since it looks largely like boilerplate. But with just three functions to generate data I think the limited duplication is fine.

PR #243, opened 2022-01-17
fixups for skipping the pki healthchecks

rcritten:
The pki healthchecks spew a lot of errors if a CA is not configured rather than gracefully reporting nothing.

A few corner cases in healthcheck skipping those were discovered and (hopefully) addressed.


antoniotorresm:
LGTM.

Issue #244, opened 2022-02-15
pylint 2.12 issues


stanislavlevin:
Pylint 2.12.2 reports several issues:
- consider-using-f-string
- redundant-u-string-prefix
- use-dict-literal
- unspecified-encoding
- use-list-literal

[pylint.log](https://github.com/freeipa/freeipa-healthcheck/files/8068544/pylint.log)

PR #245, opened 2022-02-15
Pylint 2.12

stanislavlevin:
disabled:
- `consider-using-f-string`
- `redundant-u-string-prefix`
- `use-dict-literal`
- `unspecified-encoding`
- `use-list-literal`

enabled:
- `useless-suppression`

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/244

### DIFF
--- a/.github/workflows/pipelines.yml
+++ b/.github/workflows/pipelines.yml
@@ -49,9 +49,10 @@ jobs:
                 freeipa-server \
                 freeipa-server-trust-ad \
                 tox \
-                python3-pylint \
+                python3-pip \
                 python3-pytest \
                 ; \
+            python3 -m pip install --user --ignore-installed 'pylint ~= 2.12.2' ; \
             cd /root/src; \
             tox -vv -elint; \
             "
@@ -65,7 +66,6 @@ jobs:
                 freeipa-server \
                 freeipa-server-trust-ad \
                 tox \
-                python3-pylint \
                 python3-pytest \
                 ; \
             cd /root/src; \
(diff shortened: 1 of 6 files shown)


rcritten:
Thanks for this considerable patch. The changes look valid to me, I just find some of the new linting to be annoying :-)

Particularly the open() with encoding. Is that really necessary? I'm not really the best to judge this as I'm English-only whose alphabet consists only in ASCII so I may have lucked out in the encoding assumptions. Is my assumption correct that we're only dealing with plain ASCII or is it just safer/defensive programming to be explicit about it? Can we safely ignore this instead of extending the open arguments?

stanislavlevin:
Another point is waiting for updated `pylintrc` in freeipa (I will open PR in a day or two) since `pylintrc` was force replaced with freeipa's one and thereby healthcheck inherits many skipped checks :(

rcritten:
I had read the PEP which led me to the question. It didn't really clarify things as you have. I'm ok with this. I'll hold off for your further pylint changes.

And yeah, honestly, you know pylint way better than I do hence the copying from a working, similar project.

stanislavlevin:
I synced to freeipa configuration, most of the new checkers have been skipped.

rcritten:
Just to make sure I understand, this is synced to freeipa PR https://github.com/freeipa/freeipa/pull/6196 ?

stanislavlevin:
let me recheck please.

stanislavlevin:
@rcritten, added 3 more commits to sync `pylint_plugins.py` and `pylintrc`. Now they are exactly the same as IPA's ones in today's `dev` branch.

rcritten:
Sorry for the delay, been buried with work. I'll try to get to this soon. A quick look turned up no issues.

rcritten:
LGTM, thanks for the contribution!

stanislavlevin:
@rcritten, thank you!

Issue #247, opened 2022-02-15
[IPAGroupMemberCheck] hardcoded user/group names


stanislavlevin:
New `IPAGroupMemberCheck` checker hardcoded expected user and group name as:
```python
GROUP_MEMBERS = (
    ('ipaapi', ('apache',)),
)
```
This breaks ipaplatforms having different names.

I suggest `ipaplatform.constants` instead.

PR #248, opened 2022-02-15
nss_checker: Make user/group names platform-dependent

stanislavlevin:
User/group names were hardcoded, this broke ipaplatforms having
different names.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/247
Signed-off-by: Stanislav Levin <slev@altlinux.org>

### DIFF
--- a/tests/test_ipa_nss.py
+++ b/tests/test_ipa_nss.py
@@ -7,6 +7,8 @@ from collections import namedtuple
 from unittest.mock import patch
 from util import capture_results
 
+from ipaplatform.constants import constants as platform_constants
+
 from ipahealthcheck.core import config, constants
 from ipahealthcheck.ipa.plugin import registry
 from ipahealthcheck.ipa.nss import IPAGroupMemberCheck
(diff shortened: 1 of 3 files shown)


stanislavlevin:
have to fix `tests/test_ipa_nss.py` as well.

tests/test_ipa_nss.py
```diff
@@ -22,10 +24,7 @@ def make_group(name, members):
 
 
 class TestGroupMember(BaseTest):
-    @patch('grp.getgrnam')
-    def test_ipaapi_group_ok(self, mock_grp):
-        mock_grp.return_value = make_group('apache', ('apache', 'ipaapi',))
-
+    def test_ipaapi_group_ok(self):
```
rcritten, line 27: This assumes that the test system is configured properly doesn't it?

rcritten:
Just the one question, otherwise this looks great.

tests/test_ipa_nss.py
stanislavlevin: Tests require `freeipa-server` RPM package which pulls `apache` as well, both have RPM post-scripts creating system users for them. Thus, IPA (including underlying components) has all the needed system users upon installation, the configuration is not required. So, actually mocking is required only for negative test cases. But adding a mocking for the existing group is not a big deal.

stanislavlevin:
added platform dependent changes for `files_checker`

rcritten:
PR looks good, thanks. ack.

Issue #253, opened 2022-03-01
Errors with IPADogtagCertsMatchCheck


andreasdijkman:
Running: Oracle Linux 8.5
Package-versions:
```
ipa-server-4.9.6-10.0.1.module+el8.5.0+20451+6c55862e.x86_64
ipa-healthcheck-0.7-6.module+el8.5.0+20379+1b4496cf.noarch
ipa-healthcheck-core-0.7-6.module+el8.5.0+20379+1b4496cf.noarch
```

In #193 there has been some remarks about string-replacements. I'm getting the following messages on both my IPA-nodes:
```
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPADogtagCertsMatchCheck",
    "result": "ERROR",
    "uuid": "f5f78652-4ec5-451e-b7b9-ae22123c52ef",
    "when": "20220301213649Z",
    "duration": "0.786170",
    "kw": {
      "key": "ocspSigningCert cert-pki-ca",
      "nickname": "ocspSigningCert cert-pki-ca",
      "dbdir": "/etc/pki/pki-tomcat/alias",
      "msg": "{nickname} certificate in NSS DB {dbdir} does not match entry in LDAP"
    }
  },
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPADogtagCertsMatchCheck",
    "result": "ERROR",
    "uuid": "c2ec4ec5-4a08-4c0d-8111-3976e24befe6",
    "when": "20220301213649Z",
    "duration": "0.860105",
    "kw": {
      "key": "subsystemCert cert-pki-ca",
      "nickname": "subsystemCert cert-pki-ca",
      "dbdir": "/etc/pki/pki-tomcat/alias",
      "msg": "{nickname} certificate in NSS DB {dbdir} does not match entry in LDAP"
    }
  },
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPADogtagCertsMatchCheck",
    "result": "ERROR",
    "uuid": "34c10ce3-e8a2-4b4f-b68d-c160511e335c",
    "when": "20220301213649Z",
    "duration": "0.932143",
    "kw": {
      "key": "auditSigningCert cert-pki-ca",
      "nickname": "auditSigningCert cert-pki-ca",
      "dbdir": "/etc/pki/pki-tomcat/alias",
      "msg": "{nickname} certificate in NSS DB {dbdir} does not match entry in LDAP"
    }
  },
  {
[...]

PR #254, opened 2022-03-28
Use the subject base from the IPA configuration, not REALM

rcritten:
Use the subject base from the IPA configuration, not REALM

The expected certificates were hardcoded with O={REALM} which
would return false-positives if the customer defined their
own certificate subject base.

Also add a search filter to only retrieve the certificate(s) we
want to examine rather than the entire contents.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/253


rcritten:
Thanks for the review.

PR #255, opened 2022-04-05
Allow multiple file modes in the FileChecker

rcritten:
There are some cases where a strict file mode is not
necessary. The kadmind.log is one.

It is owned root:root so there is no real difference
between 0600 and 0640. So allow both.

https://bugzilla.redhat.com/show_bug.cgi?id=2058239

### DIFF
--- a/tests/test_core_files.py
+++ b/tests/test_core_files.py
@@ -17,7 +17,8 @@ nobody = pwd.getpwnam('nobody')
 files = (('foo', 'root', 'root', '0660'),
          ('bar', 'nobody', 'nobody', '0664'),
          ('baz', ('root', 'nobody'), ('root', 'nobody'), '0664'),
-         ('fiz', ('root', 'bin'), ('root', 'bin'), '0664'),)
+         ('fiz', ('root', 'bin'), ('root', 'bin'), '0664'),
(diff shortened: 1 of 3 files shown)


flo-renaud (commented):
@rcritten thanks for the PR, LGTM.
I just have one comment re. the tests: the PR adds a new file to test (with a tuple of possible modes), but there's no assert related to this new file. I would expect something like
```
assert my_results.results[4].result == ...
assert.my_results.results[4].kw.get('msg') == ...
```
in the test_files_mode method.

tests/test_core_files.py
```diff
@@ -27,6 +28,13 @@ def make_stat(mode=33200, uid=0, gid=0):
             mode = 0660
             owner = root
             group = root
+
+       Cheat sheet equivalents:
+           0600 = 33152
+           0640 = 33184
+           0644 = 33188
+           0660 = 33200
+           0666 = 33206
```
flo-renaud, line 37: You have no idea how thankful I am for these lines! I'm sure I would have forgotten how to map things over the night :)

flo-renaud (approved):
@rcritten thanks for the update, perfect!

tests/test_core_files.py
rcritten: :-)

thanks for the review.

PR #257, opened 2022-04-28
Fixes for configuration file support for cli options

rcritten:
Three small fixes to allowing the configuration file to set command-line options discovered by QE during review process

### DIFF
--- a/src/ipahealthcheck/core/config.py
+++ b/src/ipahealthcheck/core/config.py
@@ -93,6 +93,27 @@ class Config:
             self.__d[key] = d[key]
 
 
+def convert_string(value):
+    """
+    Reading options from the configuration file will leave them as
+    strings. This breaks boolean values so attempt to convert them.
+    """
+    if not isinstance(value, str):
+        return value
+
+    if value.lower() in (
+        "true",
+        "false",
+    ):
+        return value.lower() == 'true'
+    else:
+        try:
+            value = int(value)
+        except ValueError:
+            pass
+    return value
+
+
 def read_config(config_file):
     """
     Simple configuration file reader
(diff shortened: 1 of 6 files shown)


flo-renaud:
Hi @rcritten 
thanks for the PR, the new code LGTM.
Tested the following:
- [x] add `output_type=other` in conf file, ipa-healthcheck exits with error `Unknown output-type 'other'`
- [x] add `output_type=human` in conf file, works fine
- [x] use invalid format `output_type: human` in conf file, ipa-healthcheck exits with  `Unable to parse /etc/ipahealthcheck/ipahealthcheck.conf` and prints the line with the format error
- [x] use empty value `output_type=`  in conf file, ipa-healthcheck exits with  `Empty value for output_type in /etc/ipahealthcheck/ipahealthcheck.conf [default]`
- [x] comment a setting in conf file `#output_type=human`, ipa-healthcheck doesn't take it into account and uses default JSON output type

My testing detected other issues, not related to this PR:
- [ ] add `debug=False` in the conf file, ipa-healthcheck runs in debug mode. This is because `if options.debug` on line 366 is equivalent to `if 'False'` (the test compares a string value, not a boolean).
- [ ] add `verbose=False` in the conf file, same issue as above with debug.

rcritten:
Thanks for the review. I added a function to try to convert the configuration strings into native python data types.

src/ipahealthcheck/core/config.py
```diff
@@ -93,6 +93,27 @@ def merge(self, d):
...
+    strings. This breaks boolean values so attempt to convert them.
+    """
+    if not isinstance(value, str):
+        return value
+
+    if value.lower() in (
+        "true",
+        "false",
+    ):
+        return value.lower() == 'true'
```
flo-renaud, line 108: With the convert_string function, `debug=other` still triggers the debug mode. The boolean options should probably be handled separately (with something like `if key in ('debug', 'verbose'):...`)

rcritten: I was trying to avoid that because I can't control what other users of healthcheck may configure as variables (pki has its own pki-healthcheck command). I suppose I can add it for the currently known booleans. I don't think I have access to the argparse options directly at this point so I can't auto-detect the type.

flo-renaud:
Hi @rcritten
thanks for the updated PR, LGTM.
Tested the following:

- [x] add `output_type=other` in conf file, ipa-healthcheck exits with error `Unknown output-type 'other'`
- [x] add `output_type=human` in conf file, works fine
- [x] use invalid format `output_type: human` in conf file, ipa-healthcheck exits with `Unable to parse /etc/ipahealthcheck/ipahealthcheck.conf` and prints the line with the format error
- [x] use empty value `output_type=` in conf file, ipa-healthcheck exits with `Empty value for output_type in /etc/ipahealthcheck/ipahealthcheck.conf [default]`
- [x] comment a setting in conf file `#output_type=human`, ipa-healthcheck doesn't take it into account and uses default JSON output type
- [x] add `debug=False` / `verbose=False` in the conf file, no debug/verbose output
- [x] add `debug=true` in the conf file, debug output

flo-renaud (approved):

rcritten:
Thanks for the careful review.

Issue #260, opened 2022-05-05
Don't rely on certificate serial number 1 when testing basic CA connectivity


rcritten:
ipa-healthcheck uses a simple command, `ipa cert-show 1`, to test that the CA basically works. This:

1. The CA is up and listening
2. Ensures that the IPA RA cert can authenticate
3. The CA can handle a simple request

With the addition of Random Serial Numbers there is no longer a guarantee that the CA certificate will be serial number 1.

So instead read the serial number out of the certificate and show that instead.

PR #261, opened 2022-05-05
Read the IPA CA certificate to obtain the serial number

rcritten:
The dogtag connectivity test contains the command
ipa cert-show to verify that the CA basically works
and we are allowed to use it. It had a hardcoded 1 as
the CA certificate since pre-random serial numbers it was
a predictable value.

Instead read the CA cert and pluck the serial number from
it and use that value instead.

https://github.com/freeipa/freeipa-healthcheck/issues/260


src/ipahealthcheck/dogtag/ca.py
```diff
@@ -102,10 +102,13 @@ def check(self):
             logger.debug('CA is not configured, skipping connectivity check')
             return
 
+        cert = x509.load_certificate_from_file(paths.IPA_CA_CRT)
```
flo-renaud, line 105: The file `/etc/ipa/ca.crt` may contain multiple certificates, including external ones added with `ipa-cacert-manage install`, and the `load_certificate_from_file` function reads the first one if there are multiple certs in the file. As a consequence, serial_number could correspond to a cert that is not IPA CA.
Another way could be to pick any cert from ou=authorities,ou=ca,o=ipaca (the external CAs are not added there).

I am not sure it's a big deal, given that the code handles the error 'not found', but in this case do we really need to extract a serial number instead of hardcoding "1"?

flo-renaud (commented):
Hi @rcritten 
thanks for the PR. Please find an inline comment.

src/ipahealthcheck/dogtag/ca.py
```diff
@@ -102,10 +103,40 @@ def check(self):
             logger.debug('CA is not configured, skipping connectivity check')
             return
 
+        config = api.Command.config_show()
+        subject_base = config['result']['ipacertificatesubjectbase'][0]
```
flo-renaud, line 107: When the server is installed with `--ca-subject`, the CA subject can be different from `CN=Certificate Authority,$BASE`.
The value can be found using `get_ca_subject()` from ipalib/install/certstore.py or read directly from `cn=ipa,cn=cas,cn=ca,$BASE`, in the attribute `ipaCaSubjectDN`.

src/ipahealthcheck/dogtag/ca.py
```diff
@@ -102,10 +103,40 @@ def check(self):
...
+        found = False
+        for cert in certs:
+            if DN(cert.subject) == ipa_subject:
+                found = True
+                break
+
+        if not found:
+            yield Result(self, constants.ERROR,
+                         key='ipa_ca_cert_not_found',
+                         subject=ipa_subject,
```
flo-renaud, line 129: ipa_subject needs to be cast to a str, otherwise ipa-healthcheck outputs: `Output raised TypeError: Object of type DN is not JSON serializable`

flo-renaud (commented):
Hi @rcritten 
please find inline comments.

src/ipahealthcheck/dogtag/ca.py
rcritten: It's true but both of these methods return the Subject and not the base (well, it's there implicitly). I could use string substitution, or the DN module, to drop the RDN I guess. An API call is more expensive than LDAP but gives me the value I need.

src/ipahealthcheck/dogtag/ca.py
rcritten: Thanks, nice catch. It also means I'm missing a test case.

rcritten: Updated patch to fix the DN issue and include several new tests for better coverage.

src/ipahealthcheck/dogtag/ca.py
```diff
@@ -102,10 +103,40 @@ def check(self):
             logger.debug('CA is not configured, skipping connectivity check')
             return
 
+        config = api.Command.config_show()
+        subject_base = config['result']['ipacertificatesubjectbase'][0]
+        ipa_subject = DN(('CN', 'Certificate Authority'), subject_base)
```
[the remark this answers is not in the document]
rcritten: Right, I added a call to ipaserver.install.ca.lookup_ca_subject().

rcritten:
It's true, I had a bunch of unpushed changes in my tree. Should be up-to-date now.

flo-renaud (approved):
@rcritten thanks for the update, everything works fine now.

PR #262, opened 2022-05-23
Tests: Debug option should be respected in cfg file

miskopo:
Test if the debug=True option is respected if provided
in the configuration file.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=2079861

Signed-off-by: Michal Polovka <mpolovka@redhat.com>

### DIFF
--- a/tests/test_options.py
+++ b/tests/test_options.py
@@ -43,3 +43,30 @@ def test_options_merge(mock_parse, mock_run, mock_service):
         assert run.options.indent == 5
     finally:
         os.remove(config_path)
+
+
+@patch('ipahealthcheck.core.core.run_service_plugins')
+@patch('ipahealthcheck.core.core.run_plugins')
+@patch('ipahealthcheck.core.core.parse_options')
+def test_cfg_file_debug_option(mock_parse, mock_run, mock_service):
+    """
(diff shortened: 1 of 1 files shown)


tests/test_options.py
```diff
@@ -43,3 +43,30 @@ def test_options_merge(mock_parse, mock_run, mock_service):
...
+    mock_run.return_value = Results()
+    mock_parse.return_value = options
+    fd, config_path = tempfile.mkstemp()
+    os.close(fd)
+    with open(config_path, "w") as fd:
+        fd.write('[default]\n')
+        fd.write('debug=True\n')
+
+    try:
+        run = RunChecks(['ipahealthcheck.registry'], config_path)
```
rcritten, line 67: The registry loading is finding the pki checks and there is something strange going on with initialization, I think because it runs through it twice in this test class.

An easy fix is to do this instead: `run = RunChecks([''], config_path)` for both this test and the other.

We don't need to initialize the plugins for this to show that the option has been set properly.

miskopo: Thank you, the issue is now resolved.

rcritten:
Looks good, thanks.

PR #264, opened 2022-06-01
gha: Replace F34/35 with F35/36

rcritten:
- dropping Fedora 34 because it is due to EOL soon


rcritten:
Tests are passing, merging.

Issue #148, opened 2020-08-12
Display WARNING when ipa-healtcheck tool is run as normal linux user.


menonsudhir:
Currently when a normal linux user runs healthcheck tool it gives error, instead we should display warning message such as only root user can run the tool.

[testuser@master]$ ipa-healthcheck
/usr/lib/python3.7/site-packages/ipahealthcheck/core/core.py:269: DeprecationWarning: Trying deprecated initialization API: initialize() takes 3 positional arguments but 4 were given
DeprecationWarning)
Introspect error on :1.17:/org/fedorahosted/certmonger: dbus.exceptions.DBusException: org.freedesktop.DBus.Error.AccessDenied: Sender is not authorized to send message

/usr/lib/python3.7/site-packages/ipahealthcheck/ipa/host.py:41: DeprecationWarning: Use 'ipapython.ipautil.remove_file'
installutils.remove_file(ccache_name)
[
{
"source": "ipahealthcheck.dogtag.ca",
"check": "DogtagCertsConfigCheck",
"result": "CRITICAL",
"uuid": "dffa0f74-ea29-4c4e-8a80-d65fcd708a79",
"when": "20200811122952Z",
"duration": "0.002170",
"kw": {
"exception": "NSSDB '/etc/pki/pki-tomcat/alias' not initialized."
}
},
{
"source": "ipahealthcheck.dogtag.ca",
"check": "DogtagCertsConnectivityCheck",
"result": "ERROR",
"uuid": "9c445db9-d3f3-4a34-a5ab-4b3cbe1700af",
"when": "20200811122952Z",
"duration": "0.006043",
"kw": {
"msg": "Request for certificate failed, ldap2 is not connected (ldap2_140148405442832 in MainThread)"
}
},
{
"source": "ipahealthcheck.ds.backends",
"check": "BackendsCheck",
"result": "CRITICAL",
"uuid": "ff3512e6-4483-4762-8b71-01c45973e250",
"when": "20200811122952Z",
"duration": "0.000178",
"kw": {
"exception": "Could not find configuration for instance: FEDORA31-TEST"
}
},
{
"source": "ipahealthcheck.ds.config",
"check": "ConfigCheck",
"result": "CRITICAL",
"uuid": "2ac2daf7-1fd0-4152-899b-4a326d271724",
"when": "20200811122952Z",
"duration": "0.000049",
"kw": {
"exception": "Could not find configuration for instance: FEDORA31-TEST"
}
},
{
[...]

PR #267, opened 2022-06-16
Require root to run ipa-healthcheck

rcritten:
The vast majority of checks require root access so enforce
it at run time.

This won't affect other runtimes that use healthcheck-core.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/148


flo-renaud (approved):
@rcritten 
thanks for the updated man page. LGTM.

PR #269, opened 2022-06-16
Skip AD domains with posix ranges in the catalog check

rcritten:
The catalog check is intended to ensure that the trust is
working by looking up a user. For a non-posix range we can use
the Administrator user because it has a predicible SID.

With a posix range the UID/GID may not be set so the lookup
can fail (with an empty return value).

So skip domain which have a posix range associated with it.

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1775199

### DIFF
--- a/src/ipahealthcheck/ipa/trust.py
+++ b/src/ipahealthcheck/ipa/trust.py
@@ -183,6 +183,7 @@ class IPATrustDomainsCheck(IPAPlugin):
             except Exception as e:
                 yield Result(self, constants.WARNING,
                              key='domain-status',
+                             domain=domain,
                              error=str(e),
                              msg='Execution of {key} failed: {error}')
                 continue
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/ipa/trust.py
```diff
@@ -280,20 +285,41 @@ def check(self):
...
+                                 '{error}')
                 continue
 
             if not id:
                 yield Result(self, constants.WARNING,
                              key=sid,
+                             domain=trust_domain['domain'],
                              error='returned nothing',
-                             msg='Look up of {key} {error}')
+                             msg='Look up of {key} for {domain} {error}')
```
flo-renaud, line 322: Same comment as above, the message should mention {key}-500

flo-renaud (commented):
Hi @rcritten 
thanks for the PR, LGTM. I have a minor comment re. the message but it's otherwise ok.

flo-renaud (approved):
@rcritten 
thanks for the updated patch, LGTM.

Issue #93, opened 2019-11-24
Use the DN class to compare certificate subjects


rcritten:
There can be at least a case difference with certificate subjects vs what is expected and what is issued. Use the DN class to do the comparison.

From a user report:
```
{
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPARAAgent",
    "result": "ERROR",
    "uuid": "30876792-422b-4ef0-ac6f-e8eaa4742850",
    "when": "20191121140955Z",
    "duration": "0.008288",
    "kw": {
      "expected": "2;6;CN=IPA-IPASERVER-CA;CN=IPA RA,O=IPA.CORP-DOMAIN",
      "got": "2;6;cn=IPA-IPASERVER-CA;CN=IPA RA,O=IPA.CORP-DOMAIN",
      "msg": "RA agent description does not match 2;6;cn=IPA-IPASERVER-CA;CN=IPA RA,O=IPA.CORP-DOMAIN in LDAP 
and 2;6;CN=IPA-IPASERVER-CA;CN=IPA RA,O=IPA.CORP-DOMAIN expected"
    }
```

PR #272, opened 2022-06-17
Use DN to compare agent cert issuer and subjects

rcritten:
The comparison was doing a string compare of the expected
description value.  This worked most of the time but if there
were simple case differents in the attributes that would
cause a false failure.

Instead compare them separately using the DN class to do
the comparison.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/93

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -1021,7 +1021,16 @@ def check_agent(plugin, base_dn, agent_type):
             return
         ra_desc = raw_desc[0]
         ra_certs = entry.get('usercertificate')
-        if ra_desc != description:
+        (_version, exp_serial, exp_issuer, exp_subject) = \
+            ra_desc.split(';')
+        matched = all(
+            [
+                str(serial_number) == exp_serial,
+                DN(issuer) == DN(exp_issuer),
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/ipa/certs.py
```diff
@@ -1021,7 +1021,16 @@ def check_agent(plugin, base_dn, agent_type):
...
         ra_desc = raw_desc[0]
         ra_certs = entry.get('usercertificate')
-        if ra_desc != description:
+        (_version, exp_serial, exp_issuer, exp_subject) = \
+            ra_desc.split(';')
+        matched = all(
+            [
+                str(serial_number) == exp_serial,
+                DN(issuer) == DN(exp_issuer),
+                DN(subject) == DN(exp_subject),
```
flo-renaud, line 1030: According to lines 976 and 977, issuer and subject are already of DN type so there is no need to use DN(issuer) and DN(subject) (but I agree it doesn't do any harm).

rcritten: I was being a bit paranoid here. If a value is already an RDN then the DN class simply copies it as-is so it doesn't create a lot of overhead.

Issue #270, opened 2022-06-17
IPADNSSystemRecordsCheck displays warning message for 2 expected ipa-ca AAAA records.


menonsudhir:
ipa-server and ipa-replica have been installed using ipv4 and ipv6 address, but when healthcheck test is run it displays WARNING message as below

#ipa-server-install --setup-dns --forwarder=ipv4-add -a password -p pasword -n testrealm.test -r TESTREALM.TEST --hostname server.testrealm.test --ip-address=ipv6-add --auto-reverse -U

#ipa-replica-install --setup-dns --forwarder=ipv4-add --realm TESTREALM.TEST -n testrealm.test -w password --setup-ca -P admin --ip-address=ipv4-add --server=server.testrealm.test -v --ip-address=ipv6-add -U --skip-conncheck

[root@server ~]# ipa dnsrecord-find
Zone name: testrealm.test
Record name: @
NS record: server.testrealm.test., replica.testrealm.test.

Record name: _kerberos
TXT record: "TESTREALM.TEST"
URI record: 0 100 "krb5srv:m:tcp:server.testrealm.test.", 0 100 "krb5srv:m:udp:server.testrealm.test.", 0 100 "krb5srv:m:tcp:replica.testrealm.test.", 0 100 "krb5srv:m:udp:replica.testrealm.test."

Record name: _kpasswd
URI record: 0 100 "krb5srv:m:tcp:server.testrealm.test.", 0 100 "krb5srv:m:udp:server.testrealm.test.", 0 100 "krb5srv:m:tcp:replica.testrealm.test.", 0 100 "krb5srv:m:udp:replica.testrealm.test."

Record name: _kerberos._tcp
SRV record: 0 100 88 server.testrealm.test., 0 100 88 replica.testrealm.test.

Record name: _kerberos-master._tcp
SRV record: 0 100 88 server.testrealm.test., 0 100 88 replica.testrealm.test.

Record name: _kpasswd._tcp
SRV record: 0 100 464 server.testrealm.test., 0 100 464 replica.testrealm.test.

Record name: _ldap._tcp
SRV record: 0 100 389 server.testrealm.test., 0 100 389 replica.testrealm.test.

Record name: _kerberos._udp
SRV record: 0 100 88 server.testrealm.test., 0 100 88 replica.testrealm.test.

Record name: _kerberos-master._udp
SRV record: 0 100 88 server.testrealm.test., 0 100 88 replica.testrealm.test.

[...]

PR #274, opened 2022-07-06
Use new approach to validate whether the ipa-ca DNS record is complete

rcritten:
Related to the ticket:

    Use new approach to validate whether the ipa-ca DNS record is complete
    
    The previous method counted the number of servers with CA's and
    expected an identical count of servers in ipa-ca, for each of the
    A and AAAA types.
    
    If one server had only A or AAAA records then this count could be
    off and issue a spurious warning.
    
    Instead get the list of A and AAAA records for servers with a CA
    and compare the IP addresses to those of the A and AAAA records
    of ipa-ca. Return a warning if any are missing or not expected
    (e.g. a server was removed but remains in ipa-ca).
    
    https://github.com/freeipa/freeipa-healthcheck/issues/270
    
I ran into this while testing the above patch so went ahead and included the fix:

    Use exceptions to indicate parsing errors, not a return value
    
    The validation in parse_options() retured a 1 on failure.
    Raise an exception instead and expect the caller to handle it.
    
### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -7,8 +7,9 @@ import json
 import logging
 import pkg_resources
 import signal
-import warnings
+import sys
 import traceback
+import warnings
 
 from datetime import datetime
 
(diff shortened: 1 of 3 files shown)


rcritten:
The bad ipa-ca record is not a healthcheck issue. I opened https://pagure.io/freeipa/issue/9195

rcritten:
Thanks for the review!

Issue #276, opened 2022-07-28
ipa-healthcheck doesn't support certificates stored in tokens


rcritten:
For example the certificate may be visible in the softoken but it is stored in a PKCS#11 token so lacks the private key:

  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertNSSTrust",
    "result": "ERROR",
    "uuid": "1568211a-4276-4c49-a41c-b71853027609",
    "when": "20220728182829Z",
    "duration": "0.262080",
    "kw": {
      "key": "subsystemCert cert-pki-ca",
      "expected": "u,u,u",
      "got": ",,",
      "nickname": "subsystemCert cert-pki-ca",
      "dbdir": "/etc/pki/pki-tomcat/alias",
      "msg": "Incorrect NSS trust for {nickname} in {dbdir}. Got {got} expected {expected}."
    }

rcritten:
Here it used the NSS Certificate DB token which doesn't have the private key hence the false positive.

PR #277, opened 2022-09-29
Add limited support for CA certificates on a hardware token (HSM)

rcritten:
dogtagpki supports storing its subsystem and CA certificates on an HSM. Look up the token name and password in the NSS db password file. If a token exists then include that in the lookup and expect (require) the CA, audit, ocsp and subsystem certificates to be there. If a KRA is also configured then those certificates will be in the HSM as well.

PKI supports mixing and matching but for now this only supports a simplistic one HSM or no HSM.

This requires an update to IPA where certificates can be looked up by token. At this point HSM is not yet supported in IPA but once it is then this will just work.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/276

Validating this is tricky without an HSM-enabled IPA which isn't yet available outside of my workstation... Best you can do is verify this causes no harm.

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -5,6 +5,7 @@ from __future__ import division
 
 from datetime import datetime, timezone, timedelta
 import itertools
+from inspect import signature
 import logging
 import os
 import socket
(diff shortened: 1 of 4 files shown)


rcritten:
I guess I'd hoped I could get away with the unexpected-argument in the call to db.list_certs(token=token). 

It's sort of a chicken-and-egg problem. There should be no tokens until HSM support is enabled so this should be fine, but I think I'll hold off on this PR until I can get IPA updated to at least handle tokens in ipapython/certdb.py and ipaserver/install/certs.py.

Issue #278, opened 2022-10-06
ModuleNotFoundError: No module named 'ipahealthcheck.core'


gwikleag:
On Debian with the bullseye backports repository configured I get the error below.  My guess is that _ipahealtcheck_ is designed to be run only on freeipa servers and doesn't work on ipa clients.  But where is the helpful error or documentation that says that?  Maybe I did not install it according to recommendations? Where are installation instructions?  Usually they are in the front page README on GitHub.

```
$ sudo apt install freeipa-healthcheck
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
Suggested packages:
  python3-ipaserver
The following NEW packages will be installed:
  freeipa-healthcheck
0 upgraded, 1 newly installed, 0 to remove and 0 not upgraded.
Need to get 29.7 kB of archives.
After this operation, 204 kB of additional disk space will be used.
Get:1 http://ftp.us.debian.org/debian bullseye/main amd64 freeipa-healthcheck all 0.8-1 [29.7 kB]
Fetched 29.7 kB in 7s (4,557 B/s)              
Selecting previously unselected package freeipa-healthcheck.
(Reading database ... 205509 files and directories currently installed.)
Preparing to unpack .../freeipa-healthcheck_0.8-1_all.deb ...
Unpacking freeipa-healthcheck (0.8-1) ...
Setting up freeipa-healthcheck (0.8-1) ...
Created symlink /etc/systemd/system/multi-user.target.wants/ipa-healthcheck.service → /lib/systemd/system/ipa-healthcheck.service.
Created symlink /etc/systemd/system/multi-user.target.wants/ipa-healthcheck.timer → /lib/systemd/system/ipa-healthcheck.timer.
Processing triggers for man-db (2.9.4-2) ...

$ ipa-healthcheck 
Traceback (most recent call last):
  File "/usr/local/bin/ipa-healthcheck", line 33, in <module>
    sys.exit(load_entry_point('ipahealthcheck==0.11', 'console_scripts', 'ipa-healthcheck')())
  File "/usr/local/bin/ipa-healthcheck", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/usr/lib/python3.9/importlib/metadata.py", line 77, in load
    module = import_module(match.group('module'))
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
[...]

PR #279, opened 2022-10-07
Clarify in the README that healthcheck is only for servers

rcritten:
I also mentioned that it only checks the local server, not others currently.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/278


flo-renaud (approved):
Thanks for the PR, LGTM.

Issue #285, opened 2023-03-08
error when the IPA is intialised with an extern CA


m3xiz:
At installation time, IPA provided a certificate request and this one is signed by an external CA.
The heakthcheck is producing this output:
ra.get_certificate(): Request failed with status 404: Non-2xx response from CA REST API: 404. Certificate ID 0x6596a821b68c5857 not found (404)
[
  {
    "source": "ipahealthcheck.dogtag.ca",
    "check": "DogtagCertsConnectivityCheck",
    "result": "ERROR",
    "uuid": "68600eea-1f78-4628-95f2-b3a5a8a4f8db",
    "when": "20230308105921Z",
    "duration": "0.165748",
    "kw": {
      "key": "cert_show_1",
      "error": "Certificate operation cannot be completed: Request failed with status 404: Non-2xx response from CA REST API: 404. Certificate ID 0x6596a821b68c5857 not found (404)",
      "serial": "7320223107087358039",
      "msg": "Serial number not found: {error}"
    }
  }
]

The missing certificate is the sub CA signed by the external CA. I believe this is not an error and this message should not be raised. The error is reproducable.

flo-renaud:
Hi @m3xiz 
I was also able to reproduce the issue with the following packages:
dogtag-pki-base-11.2.0-2.fc37.noarch
freeipa-server-4.10.1-1.fc37.x86_64
freeipa-healthcheck-0.12-1.fc37.noarch

The code is internally calling a command equivalent to `ipa cert-show <serial for IPA CA>`, which ends up calling PKI rest API (`GET /ca/rest/certs/<serial>`) and PKI returns an error `Certificate ID <serial> not found`. We would need to check with PKI team if IPA CA should be visible through this call even when it's an externally-signed CA.

rcritten:
It pulls the serial number of the root CA and tries to do a cert-show which will likely fail, or worse be a false positive, as the certificate doesn't exist in the local CA because it was issued externally.

I think we should just skip this check when the CA issuer doesn't match the subject, as suggested.

rcritten:
https://github.com/freeipa/freeipa-healthcheck/pull/286

PR #286, opened 2023-03-27
Don't error in DogtagCertsConnectivityCheck with external CAs

rcritten:
The serial number of externally-signed CA's were being checked using ipa-cert-find which invariably would return a not found error.

Instead return SUCCESS to effectively skip the check.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/285

### DIFF
--- a/src/ipahealthcheck/dogtag/ca.py
+++ b/src/ipahealthcheck/dogtag/ca.py
@@ -12,10 +12,8 @@ from ipahealthcheck.core import constants
 from ipalib import api, errors, x509
 from ipaplatform.paths import paths
 from ipaserver.install import certs
-from ipaserver.install import ca
 from ipaserver.install import krainstance
 from ipapython.directivesetter import get_directive
-from ipapython.dn import DN
 from cryptography.hazmat.primitives.serialization import Encoding
 
 logger = logging.getLogger()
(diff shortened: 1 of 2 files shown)


flo-renaud (commented):
Hi @rcritten 
if IPA is installed with embedded CA, the connectivity check can use the RA cert in all the cases (self-signed or externally signed CA), as this cert always exists and can be found with PKI API.

rcritten:
Excellent observation! Implemented.

flo-renaud (commented):
Hi @rcritten 
thanks for the update. Minor nitpicks, not related to this patch but that could be fixed with it:
- line 141, the message is missing a closing }
- lines 126, 132 and 138, we could replace the key 'cert_show_1' with 'cert_show_ra'

I tested the following scenarios with an externally-signed CA:
- missing RA cert file
- RA cert file contains a cert inconsistent with what is stored in LDAP
- pki stopped
All the above issues were properly reported.

rcritten:
> thanks for the update. Minor nitpicks, not related to this patch but that could be fixed with it:
> 
>     * line 141, the message is missing a closing }

Fixed.

>     * lines 126, 132 and 138, we could replace the key 'cert_show_1' with 'cert_show_ra'

Done. I modified the test expectations as well, including the typical serial number for the RA (it doesn't really matter in the end but non-zero is more obvious)

flo-renaud (approved):
@rcritten 
thanks for the update, works for me. I tested the following scenarios with an externally-signed CA:

- missing RA cert file
- RA cert file contains a cert inconsistent with what is stored in LDAP
- pki stopped

All the above issues were properly reported.

Issue #284, opened 2023-02-07
IPADNSSystemRecordsCheck's false positive result for ipa-ca rrs


stanislavlevin:
`freeipa-healthcheck` 0.12 reports false positive result for `ipa-ca` rrs in case they are missing.

Discovered by `TestIpaHealthCheckWithoutDNS::test_ipa_dns_systemrecords_check`
https://github.com/freeipa/freeipa/blob/master/ipatests/test_integration/test_ipahealthcheck.py#L1612
being run in azure pipelines.

how to reproduce:
- point nameserver to 8.8.8.8 in `/etc/resolv.conf`
- install ipa-server without embedded dns
- run ipa-healthcheck

output:
```console
[root@master1 /]# ipa-healthcheck --debug --source ipahealthcheck.ipa.idns --check IPADNSSystemRecordsCheck --output-type json --all
...
Calling check <ipahealthcheck.ipa.idns.IPADNSSystemRecordsCheck object at 0x7f443a6a1060>
raw: server_find(None, version='2.251', no_members=False, servrole='IPA master')
server_find(None, all=False, raw=False, version='2.251', no_members=False, pkey_only=False, servrole=('IPA master',))
raw: server_role_find(None, server_server=None, role_servrole='IPA master', status='enabled', include_master=True, version='2.251')
server_role_find(None, server_server=None, role_servrole='IPA master', status='enabled', include_master=True, all=False, raw=False, version='2.251')
raw: topologysuffix_find(None, all=True, raw=True, version='2.251')
topologysuffix_find(None, all=True, raw=True, version='2.251', pkey_only=False)
raw: server_role_find(None, server_server='master1.ipa.test', status='enabled', include_master=True, version='2.251')
server_role_find(None, server_server='master1.ipa.test', status='enabled', include_master=True, all=False, raw=False, version='2.251')
Name master1.ipa.test. resolved to {UnsafeIPAddress('fe80::42:acff:fe15:3'), UnsafeIPAddress('2001:db8:1:1::3'), UnsafeIPAddress('172.21.0.3')}
Invalid IP address fe80::42:acff:fe15:3 for master1.ipa.test.: cannot use link-local IP address fe80::42:acff:fe15:3
Search DNS for SRV record of _ldap._tcp.ipa.test.
DNS record not found: NXDOMAIN
Search DNS for SRV record of _kerberos._tcp.ipa.test.
DNS record not found: NXDOMAIN
[...]

PR #287, opened 2023-03-27
Report when all ipa-ca records are missing in IPADNSSystemRecordsCheck

rcritten:
If no DNS records were returned at all then the check for mismatches was missed. Add a special case for this scenario by using a placeholder for the IP address for a given CA server. If no records are returned at all this case will catch it. If any exist at all then the current code will handle it.

This is is easily reproduced using a non-IPA DNS server like Google or Cloudflare where all lookups will fail.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/284


rcritten:
Thanks!

Issue #197, opened 2021-05-03
 ipa-healthcheck --indent with large values causes 'Output raised MemoryError: '


celestian:
Ticket was cloned from Red Hat Bugzilla (product Red Hat Enterprise Linux 8): Bug https://bugzilla.redhat.com/show_bug.cgi?id=1774037

```
Description of problem: ipa-healthcheck --indent with large values causes 'Output raised MemoryError: '


Version-Release number of selected component (if applicable):
[root@master ~]# rpm -q ipa-server
ipa-server-4.8.2-1.module+el8.2.0+4697+7171660c.x86_64
[root@master ~]# rpm -q ipa-server ipa-healthcheck
ipa-server-4.8.2-1.module+el8.2.0+4697+7171660c.x86_64
ipa-healthcheck-0.3-4.module+el8.1.0+4098+f286395e.noarch


How reproducible: Always

Steps to Reproduce:
1. Install ipa-server and ipa-healthcheck
2. Run the command with large values e.g

[root@master ~]# ipa-healthcheck --indent 1000000

Actual results:
[root@master ~]# ipa-healthcheck --indent 1000000
Output raised MemoryError: 

Expected results: indent values should have max values allowed so that MemoryError is not seen.
```

PR #288, opened 2023-03-28
Restrict the length of JSON output indent to 32

rcritten:
Too big a value and the system will exhause memory. Normally I trust users not to do things like but it doesn't hurt to constrain sometimes.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/197

### DIFF
--- a/src/ipahealthcheck/core/output.py
+++ b/src/ipahealthcheck/core/output.py
@@ -2,6 +2,7 @@
 # Copyright (C) 2019 FreeIPA Contributors see COPYING for license
 #
 
+from argparse import ArgumentTypeError
 from datetime import datetime
 import json
 import sys
@@ -14,6 +15,26 @@ class OutputRegistry(Registry):
 
 
 output_registry = OutputRegistry()
+
+
+class Int:
+    def __init__(self, minimum=0, maximum=100):
+        self.minimum = minimum
+        self.maximum = maximum
+
+    def __call__(self, arg):
+        try:
+            value = int(arg)
+        except ValueError:
+            raise ArgumentTypeError("'%s' is not an integer" % arg)
+
+        if (value < self.minimum) or (value > self.maximum):
+            raise ArgumentTypeError(
+                "'%s' is not in the range %s-%s"
+                % (value, self.minimum, self.maximum)
+            )
+
+        return value
 
 
 class Output:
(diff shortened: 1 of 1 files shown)


src/ipahealthcheck/core/output.py
```diff
@@ -16,6 +17,26 @@ class OutputRegistry(Registry):
 output_registry = OutputRegistry()
 
 
+class Int:
+    def __init__(self, minimum=0, maximum=0):
```
flo-renaud, line 21: The default value maximum=0 currently does not cause any issue because the only use of this class is setting maximum=32. But for future uses, creating an Int() without parameters would result in only the 0 value allowed. By default it would make more sense to have default maximum=a non-null value.
A check to ensure that minimum < maximum could be added.

rcritten: Yeah I guess. I like to give folks a lot of rope to hang themselves with but I can set a reasonable minimum, I just don't know what this might be used for in the future so whatever I choose will be wrong. But it doesn't have to be a gotcha  I guess.

flo-renaud (approved):
@rcritten 
thanks for the PR, works for me.

Tested with out of range /invalid values:
```
# ipa-healthcheck --indent -1
usage: ipa-healthcheck [-h] [--config CONFIG] [--verbose] [--debug] [--list-sources] [--source SOURCE] [--check CHECK]
                       [--output-type {json,human,prometheus}] [--output-file OUTPUT_FILE] [--version] [--indent INDENT]
                       [--metric-prefix METRIC_PREFIX] [--input-file INFILE] [--failures-only] [--all] [--severity {SUCCESS,WARNING,ERROR,CRITICAL}]
ipa-healthcheck: error: argument --indent: '-1' is not in the range 0-32
# ipa-healthcheck --indent 33
usage: ipa-healthcheck [-h] [--config CONFIG] [--verbose] [--debug] [--list-sources] [--source SOURCE] [--check CHECK]
                       [--output-type {json,human,prometheus}] [--output-file OUTPUT_FILE] [--version] [--indent INDENT]
                       [--metric-prefix METRIC_PREFIX] [--input-file INFILE] [--failures-only] [--all] [--severity {SUCCESS,WARNING,ERROR,CRITICAL}]
ipa-healthcheck: error: argument --indent: '33' is not in the range 0-32
# ipa-healthcheck --indent a
usage: ipa-healthcheck [-h] [--config CONFIG] [--verbose] [--debug] [--list-sources] [--source SOURCE] [--check CHECK]
                       [--output-type {json,human,prometheus}] [--output-file OUTPUT_FILE] [--version] [--indent INDENT]
                       [--metric-prefix METRIC_PREFIX] [--input-file INFILE] [--failures-only] [--all] [--severity {SUCCESS,WARNING,ERROR,CRITICAL}]
ipa-healthcheck: error: argument --indent: 'a' is not an integer

```

Issue #125, opened 2020-05-01
Detect if KRA is installed and not the CA renewal master


rcritten:
Perhaps for cluster checking.

It's possible inside a cluster to have the renewal master not have the KRA role installed which means those certificates will never be renewed. Detect this condition and ERROR on it.

rcritten:
Removing cluster flag. We can determine this based on the server roles and whether the renewal server has a KRA or not.

PR #290, opened 2023-03-30
If there are KRAs, ensure the renewal server is one

rcritten:
If there are KRAs in the topology and there isn't one on the renewal server then the KRA certificates will not be renewed because they expect another server to do it for them.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/125

### DIFF
--- a/src/ipahealthcheck/ipa/roles.py
+++ b/src/ipahealthcheck/ipa/roles.py
@@ -10,6 +10,7 @@ from ipahealthcheck.core.plugin import duration
 from ipahealthcheck.core import constants
 
 from ipalib import api
+from ipaserver.install import krainstance
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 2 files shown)


flo-renaud (commented):
Hi @rcritten 
Thanks for the PR. The new check works as expected, but I was wondering why it's performing a search to find the KRAs. The call to config_show already gathers all the needed information:
```
# ipa config-show
[...]
  IPA CA renewal master: server.ipa.test
  IPA KRA servers: replica.ipa.test
[...]
```
Any reason to not rely on this info?

rcritten:
> Any reason to not rely on this info?

No, I totally missed this. Dropped the search as it is much simpler code.

flo-renaud:
Thanks for the update, LGTM. I tested the same scenario as the pytest:
- host is not renewal master
- host is renewal master, no KRA at all
- host is renewal master, there is a KRA but not on this host
- host is renewal master and has KRA

flo-renaud (approved):

Issue #123, opened 2020-03-19
[RFE] Check for certs in stuck status


yrro:
I noticed one of the certs on a CentOS 8 server got stuck:

```
Request ID '20200123083218':
        status: NEWLY_ADDED_NEED_KEYINFO_READ_PIN
        stuck: yes
        key pair storage: type=FILE,location='/var/lib/ipa/private/httpd.key'
        certificate: type=FILE,location='/var/lib/ipa/certs/httpd.crt'
        CA: IPA
        issuer: CN=Certificate Authority,O=IPA.EXAMPLE.COM
        subject: CN=ipa2.ipa.example.com,O=IPA.EXAMPLE.COM
        expires: 2021-12-19 18:32:27 UTC
        dns: ipa2.ipa.example.com
        principal name: HTTP/ipa2.ipa.example.com@IPA.EXAMPLE.COM
        key usage: digitalSignature,nonRepudiation,keyEncipherment,dataEncipherment
        eku: id-kp-serverAuth,id-kp-clientAuth
        pre-save command:
        post-save command: /usr/libexec/ipa/certmonger/restart_httpd
        track: yes
        auto-renew: yes
```

`ipahealthcheck.ipa.certs` doesn't complain about this. It would be useful if the module checked for the `stuck` status of each certificate request.

yrro:
I guess that if I didn't fix this problem then `ipa-healthcheck` would have notified me 28 days before the cert expired so this isn't strictly necessary...

rcritten:
Thanks for the suggestion, I think this has value even if another check would eventually notice.

yrro:
Yes - the user gets a notification immediately rather than 28 days before expiry.

rcritten:
In this particular case I guess I'm surprised that an issue about unexpected cert tracking and a missing cert wasn't raised since the pinfile wasn't included. I'll also look at that code, perhaps the pin isn't considered in that check.

BornTKill:
Dear,

I also had an issue with "stuck" certificat.
Request ID '20200713131458':
	status: NEWLY_ADDED_NEED_KEYINFO_READ_PIN
	stuck: yes
	key pair storage: type=NSSDB,location='/etc/pki/pki-tomcat/alias',nickname='auditSigningCert cert-pki-ca'
	certificate: type=NSSDB,location='/etc/pki/pki-tomcat/alias',nickname='auditSigningCert cert-pki-ca'
[...]

PR #291, opened 2023-03-31
Report certmonger requests that are in the stuck state

rcritten:
These may be caught already by other checks if the tracking is configured incorrectly but it's a belt-and-suspenders approach to ensure that the certificates have been issued properly.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/123

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -1418,3 +1418,22 @@ class IPACAChainExpirationCheck(IPAPlugin):
                              path=paths.IPA_CA_CRT,
                              key=subject,
                              days=(dt - now).days)
+
+
+@registry
+class CertmongerStuckCheck(IPAPlugin):
+    """Check for certonger requests in the stuck state
+    """
+
+    @duration
+    def check(self):
+        requests = certmonger._get_requests({'stuck': True})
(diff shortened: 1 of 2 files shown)


flo-renaud (approved):
Hi @rcritten 
thanks for the new check, WFM.
I tested with the following scenario:
- remove the PIN file for the dirsrv certificate: `mv /etc/dirsrv/slapd-IPA-TEST/pwdfile.txt /etc/dirsrv/slapd-IPA-TEST/pwdfile.txt.sav`
- renew the dirsrv certificate: `getcert resubmit -i 20230407091205`
- check the request is stuck: `getcert list -i 20230407091205 | grep stuck` returns `stuck: yes`
- run the check:
 ```
# ipa-healthcheck --source ipahealthcheck.ipa.certs --check CertmongerStuckCheck
[
  {
    "check": "CertmongerStuckCheck",
    "result": "WARNING",
    "uuid": "d32781d5-09e7-4d52-89f1-57eca541e9e4",
    "when": "20230407120255Z",
    "duration": "0.019584",
    "kw": {
      "key": "20230407091205",
      "msg": "certmonger request {key} is in the stuck state"
    }
  }
]
```
- fix the PIN file, resubmit, re-run the check: no failure detected.

PR #293, opened 2023-05-27
output: fix prometheus output pluging to comply with format spec

### LINKED ISSUE #292: prometheus output plugin violates format spec
the prometheus output plugin renders HELP and TYPE annotations
without the required '#' prefix, which is required according
to the [format specification](https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format)

the problem can be reproduced by running the following command:
`ipa-healthcheck --all --output-type prometheus`

### expected output

```
# HELP ipa_healthcheck Number of healthchecks with a certain result
# TYPE ipa_healthcheck gauge
ipa_healthcheck{result="SUCCESS"} 1.0
```

### actual output

```
HELP ipa_healthcheck Number of healthchecks with a certain result
TYPE ipa_healthcheck gauge
ipa_healthcheck{result="SUCCESS"} 1.0
```

UiP9AV6Y:
use comment syntax for HELP and TYPE annotations, as specified by the prometheus test-base exposition format [1].

this change also introduces a output sub-class to reduce code duplication.

closes #292

[1] https://prometheus.io/docs/instrumenting/exposition_formats/#text-based-format

### DIFF
--- a/src/ipahealthcheck/core/output.py
+++ b/src/ipahealthcheck/core/output.py
@@ -174,6 +174,43 @@ class Human(Output):
 class Prometheus(Output):
     """Render results as Prometheus text metric exposition format"""
 
+    class Renderer:
+        """Line-based Metrics Exposition Format renderer"""
+
+        def __init__(self, metric_prefix=""):
+            self.metric_prefix = metric_prefix
+            self.output = []
+
+        def add_info(self, metric_name, metric_type=None, help=None):
+            if help is not None:
+                line = '# HELP %s_%s %s' % (
+                    self.metric_prefix, metric_name, help,
+                )
+                self.output.append(line)
+
+            if metric_type is not None:
+                line = '# TYPE %s_%s %s' % (
+                    self.metric_prefix, metric_name, metric_type,
+                )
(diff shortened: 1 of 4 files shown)


rcritten:
I think this looks good generally. It is more conformant with the spec.

I do have some minor nitpicks.

Is there a reason to convert all single quotes to double quotes? I imagine your editor did this automatically?

Also, can you add a sign-off to the commit (the -s option to git)

src/ipahealthcheck/core/output.py
```diff
@@ -174,9 +174,46 @@ def generate(self, data):
...
+        def render(self):
+            self.output.append("")
+
+            return "\n".join(self.output)
+
     options = (
-        ('--metric-prefix', dict(dest='metric_prefix', default='ipa',
-         help='Metric name prefix')),
+           ('--metric-prefix', dict(dest='metric_prefix', default='ipa',
+            help='Metric name prefix')),
```
rcritten, line 216: Why go from 4 to 8 indention?

rcritten:
Thanks for the changes, this LGTM.

PR #294, opened 2023-05-30
gha: Replace F35/36 with F37/38

rcritten:
F35 and F36 are not both obsolete. Switch to current releases.


flo-renaud (approved):
@rcritten thanks for the PR, LGTM.

Issue #298, opened 2023-08-10
Python 3.12 deprecation warning about datetime.utcnow()


flo-renaud:
With python 3.12, the function datetime.datetime.utcnow() is deprecated. Running pytest on freeipa-healthcheck  project produces warnings:
```
tests/test_system_filesystemspace.py: 3 warnings
  /root/freeipa-healthcheck/src/ipahealthcheck/core/plugin.py:16: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    start = datetime.utcnow()

[...]
tests/test_system_filesystemspace.py: 30 warnings
  /root/freeipa-healthcheck/src/ipahealthcheck/core/plugin.py:135: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    self.when = when or generalized_time(datetime.utcnow())

[...]
tests/test_system_filesystemspace.py: 30 warnings
  /root/freeipa-healthcheck/src/ipahealthcheck/core/plugin.py:19: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    end = datetime.utcnow()

tests/test_ipa_certfile_expiration.py::TestIPACertificateFile::test_certfile_expiration
  /root/freeipa-healthcheck/tests/test_ipa_certfile_expiration.py:43: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    cert = IPACertificate(not_valid_after=datetime.utcnow() +

tests/test_ipa_certfile_expiration.py::TestIPACertificateFile::test_certfile_expiration_warning
  /root/freeipa-healthcheck/src/ipahealthcheck/ipa/certs.py:381: DeprecationWarning: datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.now(datetime.UTC).
    now = datetime.utcnow()

tests/test_ipa_certfile_expiration.py::TestIPACertificateFile::test_certfile_expiration_expired
[...]

PR #299, opened 2023-08-10
Python 3.12: utcnow function is deprecated

flo-renaud:
ipa-healthcheck on python 3.12 uses datetime.utcnow() which is deprecated and produces warnings.
Replace with datetime.now(tz=UTC)

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/298

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -11,7 +11,7 @@ import sys
 import traceback
 import warnings
 
-from datetime import datetime
+from datetime import datetime, UTC
 
 from ipahealthcheck.core.config import read_config
 from ipahealthcheck.core.exceptions import TimeoutError
@@ -49,7 +49,7 @@ def run_plugin(plugin, available=(), timeout=constants.DEFAULT_TIMEOUT):
         raise TimeoutError('Request timed out')
 
     # manually calculate duration when we create results of our own
-    start = datetime.utcnow()
+    start = datetime.now(tz=UTC)
     signal.signal(signal.SIGALRM, signal_handler)
     signal.alarm(timeout)
     try:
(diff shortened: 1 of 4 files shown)


flo-renaud:
In order to see the deprecation warnings: run pytest on freeipa-healthcheck repo as described in https://github.com/freeipa/freeipa-healthcheck#readme

In order to see the exception reported by `ipa-healthcheck --source ipahealthcheck.ipa.certs --checkIPACertfileExpirationCheck`, install freeipa-server built from the master branch (containing the commit https://github.com/freeipa/freeipa/commit/09497d2df0fbd4bb5ad798e5c0798a0faa632f11)

We will need to synchronize and release a new version of freeipa and freeipa-healthcheck.

rcritten:
The issue runs in both directions. With updated healthcheck and old IPA the same issue appears: TypeError: can't compare offset-naive and offset-aware datetimes

If I align the healthcheck and IPA releases that will resolve the issue but potentially make backports difficult and I'd have to switch to IPA-aligned branching. Thus far one codebase has been able to cover 4.9, 4.10 and 4.11(beta).

A datetime object has a replace method to set a TZ and there is no offset include (in my brief testing) doing it this way:

notafter = notafter.replace(tzinfo=UTC)

I wonder if we can create a function, time_less_than, or something which will compare two times and set one or both to UTC if there is no tzinfo in the object. I think that should be universal.

flo-renaud:
@rcritten very relevant point indeed... I updated my patch so that the new ipa-healthcheck would be compatible with old and new IPA versions (IPACertificate can return naive or timezone-aware values for cert.not_valid_after and cert.not_valid_before).

rcritten:
Thanks. The change tests out ok on a non Python 3.12 system. I'll fire up one that has 3.12 and double-check it but the code looks good.

rcritten:
Ack. It works in rawhide as well.

Issue #300, opened 2023-09-28
Traceback when running ipa-healthcheck if LDAP server is down


flo-renaud:
With ipa-healthcheck 0.14, ipa-healthcheck reports a traceback if the LDAP server is not running:
```
# rpm -qa freeipa-healthcheck
freeipa-healthcheck-0.14-1.fc38.noarch
# systemctl stop dirsrv@IPA-TEST.service
# ipa-healthcheck 
Traceback (most recent call last):
  File "/usr/lib/python3.11/site-packages/ipapython/ipaldap.py", line 1088, in error_handler
    yield
  File "/usr/lib/python3.11/site-packages/ipapython/ipaldap.py", line 1246, in external_bind
    self.conn.sasl_interactive_bind_s(
  File "/usr/lib64/python3.11/site-packages/ldap/ldapobject.py", line 270, in sasl_interactive_bind_s
    return self._ldap_call(self._l.sasl_interactive_bind_s,who,auth,RequestControlTuples(serverctrls),RequestControlTuples(clientctrls),sasl_flags)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/site-packages/ldap/ldapobject.py", line 128, in _ldap_call
    result = func(*args,**kwargs)
             ^^^^^^^^^^^^^^^^^^^^
ldap.SERVER_DOWN: {'result': -1, 'desc': "Can't contact LDAP server", 'errno': 111, 'ctrls': [], 'info': 'Connection refused'}

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/bin/ipa-healthcheck", line 33, in <module>
    sys.exit(load_entry_point('ipahealthcheck==0.14', 'console_scripts', 'ipa-healthcheck')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ipahealthcheck/core/main.py", line 54, in main
    sys.exit(ipachecks.run_healthcheck())
[...]

PR #301, opened 2023-09-28
Add a dirsrv requires to services that look up their names in LDAP

rcritten:
Some services aren't included in ipaplaform.knownservices like smb and winbind. It is possible to discover the service name using the same method used by ipactl, via roles.

If dirsrv isn't started then this will blow up spectacularly so use requires as a guard against it.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/300

### DIFF
--- a/src/ipahealthcheck/meta/services.py
+++ b/src/ipahealthcheck/meta/services.py
@@ -118,6 +118,8 @@ class httpd(IPAServiceCheck):
 
 @registry
 class ipa_custodia(IPAServiceCheck):
+    requires = ('dirsrv',)
+
     def check(self, instance=''):
         self.service_name = self.get_service_name('KEYS')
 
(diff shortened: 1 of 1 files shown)


flo-renaud (approved):
@rcritten 
thanks for the PR, works for me:
```
(venv) [root@server freeipa-healthcheck]# systemctl stop dirsrv@IPA-TEST.service
(venv) [root@server freeipa-healthcheck]# ipa-healthcheck 
[
  {
    "source": "ipahealthcheck.meta.services",
    "check": "dirsrv",
    "result": "ERROR",
    "uuid": "1466294d-c561-4adf-9443-e19308f2427c",
    "when": "20231003085001Z",
    "duration": "0.005170",
    "kw": {
      "status": false,
      "msg": "dirsrv: not running"
    }
  }
]
```

Issue #302, opened 2023-09-28
module 'datetime' has no attribute 'UTC'


rcritten:
Sister issue to https://pagure.io/freeipa/issue/9454

In commit e05903dc95be0182eac310f5a9d593d93fc43fa8 we stopped using utcnow() because it is deprecated and used datatime.UTC instead. This was introduced in python 3.11 so doesn't work with older releases.

Use datetime.timezone.utc instead which is more universal. datetime.UTC is an alias to datetime.timezone.utc.

PR #303, opened 2023-09-28
Use timezone.utc instead of datetime.UTC for backwards compatibility

rcritten:
We switched to datetime.UTC because datetime.utcnow() was deprecated. This is only available in python 3.11+. Use datetime.timezone.utc instead which is available from python 3.2+

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/302

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -11,7 +11,7 @@ import sys
 import traceback
 import warnings
 
-from datetime import datetime, UTC
+from datetime import datetime, timezone
 
 from ipahealthcheck.core.config import read_config
 from ipahealthcheck.core.exceptions import TimeoutError
@@ -49,7 +49,7 @@ def run_plugin(plugin, available=(), timeout=constants.DEFAULT_TIMEOUT):
         raise TimeoutError('Request timed out')
 
     # manually calculate duration when we create results of our own
-    start = datetime.now(tz=UTC)
+    start = datetime.now(tz=timezone.utc)
     signal.signal(signal.SIGALRM, signal_handler)
     signal.alarm(timeout)
     try:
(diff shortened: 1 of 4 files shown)


flo-renaud:
Hi @rcritten 
thanks for the patch, works well for me. I tested on fedora 38 with python 3.11 and on RHEL 9.3 with python 3.9.

The patch is missing an update for `tests/test_ipa_certfile_expiration.py` though, which contains a few occurrences of UTC:
```
# pytest
============================= test session starts ==============================
platform linux -- Python 3.9.18, pytest-7.4.2, pluggy-1.3.0
rootdir: /root/freeipa-healthcheck
collected 188 items / 1 error                                                  

==================================== ERRORS ====================================
____________ ERROR collecting tests/test_ipa_certfile_expiration.py ____________
ImportError while importing test module '/root/freeipa-healthcheck/tests/test_ipa_certfile_expiration.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/lib64/python3.9/importlib/__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/test_ipa_certfile_expiration.py:18: in <module>
    from datetime import datetime, timedelta, UTC
E   ImportError: cannot import name 'UTC' from 'datetime' (/usr/lib64/python3.9/datetime.py)
=========================== short test summary info ============================
ERROR tests/test_ipa_certfile_expiration.py
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 0.60s ===============================
```

rcritten:
> Hi @rcritten thanks for the patch, works well for me. I tested on fedora 38 with python 3.11 and on RHEL 9.3 with python 3.9.
> 
> The patch is missing an update for `tests/test_ipa_certfile_expiration.py` though, which contains a few occurrences of UTC:

Can't believe I missed those. Fixed now.

Issue #307, opened 2023-10-27
Support verifying LWCA (SubCA) certmonger requests


rcritten:
healthcheck doesn't currently expect to see tracked LWCA certificates. These typically have the form of:

```
Request ID '20231027130916':
        status: MONITORING
        stuck: no
        key pair storage: type=NSSDB,location='/etc/pki/pki-tomcat/alias',nickna
me='caSigningCert cert-pki-ca 9a93fc7f-4246-4986-af45-7d7f85dd136b',token='NSS C
ertificate DB',pin set
        certificate: type=NSSDB,location='/etc/pki/pki-tomcat/alias',nickname='c
aSigningCert cert-pki-ca 9a93fc7f-4246-4986-af45-7d7f85dd136b',token='NSS Certif
icate DB'
        CA: dogtag-ipa-ca-renew-agent
        issuer: CN=Certificate Authority,O=EXAMPLE.TEST
        subject: CN=test
        issued: 2023-10-27 09:07:04 EDT
        expires: 2043-10-27 08:59:05 EDT
        key usage: digitalSignature,nonRepudiation,keyCertSign,cRLSign
        profile: caCACert
        pre-save command: /usr/libexec/ipa/certmonger/stop_pkicad
        post-save command: /usr/libexec/ipa/certmonger/renew_ca_cert "caSigningC
ert cert-pki-ca 9a93fc7f-4246-4986-af45-7d7f85dd136b"
        track: yes
        auto-renew: yes
```

They are currently reported as a warning about an unknown tracked certificate.

We can fetch the CA UUID's from LDAP and build a template request for them so they no longer warn.

rcritten:
The failure looks like:

```
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPACertTracking",
    "result": "WARNING",
    "uuid": "4d31e52d-0da7-4434-8da8-2fc4bcae5e25",
    "when": "20231027151337Z",
    "duration": "1.124710",
    "kw": {
      "key": "20231027130916",
      "msg": "certmonger tracking request {key} found and is not expected on an IPA master."
    }
  }
```

rcritten:
IPA does not generate a tracking request for subCA's after creation. One needs to run ipa-server-upgrade to add them.

PR #308, opened 2023-10-27
Support validating LWCA certmonger requests

rcritten:
The LWCA ids are UUID4 format and are stored in LDAP so we can retrieve the list (ignoring the ipa entry) and construct what the request should look like.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/307

### DIFF
--- a/src/ipahealthcheck/core/core.py
+++ b/src/ipahealthcheck/core/core.py
@@ -10,6 +10,7 @@ import signal
 import sys
 import traceback
 import warnings
+from ipalib import api
 
 from datetime import datetime, timezone
 
@@ -464,6 +465,7 @@ class RunChecks:
             return 1
 
         results = exclude_keys(config, results)
+        api.Backend.ldap2.disconnect()
 
         try:
             output.render(results)
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -1348,8 +1348,7 @@ class IPACertRevocation(IPAPlugin):
 
             # Now we have the cert either way, check the recovation
             try:
-                result = api.Command.cert_show(cert.serial_number,
-                                               all=True)
+                result = api.Command.cert_show(cert.serial_number)
             except Exception as e:
                 yield Result(self, constants.ERROR,
                              key=id,
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -35,6 +35,8 @@ from ipapython.ipaldap import realm_to_serverid
 logger = logging.getLogger()
 DAY = 60 * 60 * 24
 
+expected_requests = []  # to cache the expected certmonger requests
+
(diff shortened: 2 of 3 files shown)


rcritten:
While developing the original patch I noticed that cn=cas was being hammered. It was because the call to cert_find included all=True which was confusing the cache. We in fact don't need all attributes. This increased the cache hits from 7 to 24.

rcritten:
Switched from caching only the LWCA requests to all expected requests. This saves a fair bit of effort and a few potential LDAP calls.

flo-renaud (commented):
Hi @rcritten 
thanks for the new check, LGTM. I tested the following:
- [x] fresh install without any LWCA. Prints a warning that there is no LWCA entry:
```
# ipa-healthcheck --source ipahealthcheck.ipa.certs 
Search for LWCA entries failed: no matching entry found
[]
```
Maybe the message could be made a bit less negative since it's a completely valid situation, for instance: `Skipping LWCA checks because the search did not return any LWCA entry`

- [x] Add a LWCA with `ipa ca-add subca --subject cn=subca,O=IPA.TEST` but do not run ipa-certupdate. The subca entry is found but there is no tracking yet:
```
# ipa-healthcheck --source ipahealthcheck.ipa.certs 
[
  {
    "check": "IPACertTracking",
    "result": "ERROR",
    "uuid": "c0dbcc12-67d0-4f40-a7ee-e9db42168cda",
    "when": "20231106094520Z",
    "duration": "0.406647",
    "kw": {
      "key": "cert-database=/etc/pki/pki-tomcat/alias, cert-nickname=caSigningCert cert-pki-ca f4315412-6c51-4576-ab12-2a3088720869, ca-name=dogtag-ipa-ca-renew-agent, cert-presave-command=/usr/libexec/ipa/certmonger/stop_pkicad, cert-postsave-command=/usr/libexec/ipa/certmonger/renew_ca_cert \"caSigningCert cert-pki-ca f4315412-6c51-4576-ab12-2a3088720869\", template-profile=caCACert",
      "msg": "Expected certmonger tracking is missing for {key}. Automated renewal will not happen for this certificate"
    }
  }
]
```
- [x] run `ipa-certupdate` to create the tracking request. No error reported
```
# ipa-healthcheck --source ipahealthcheck.ipa.certs 
[]
```
- [x] disable and remove the ca (`ipa ca-disable subca` and `ipa ca-del subca`, remove the tracking request with `getcert stop-tracking -i ID` and run `ipa-certupdate`. No error reported:
```
# ipa-healthcheck --source ipahealthcheck.ipa.certs 
Search for LWCA entries failed: no matching entry found
[]
```

rcritten:
Huh. I'd have sworn I was logging the LWCA search result as debug and not warn. I'll fix that up and drop the "failed" bit. The search succeeded but returned nothing.

flo-renaud (approved):
@rcritten thanks for the updated patch, works for me.

Issue #310, opened 2023-11-09
Traceback when calling pki-healthcheck


flo-renaud:
With the new 0.15 release of freeipa-healthcheck, the command pki-healthcheck reports a traceback that didn't happen with 0.14.
Successful execution with 0.14:
```
[root@server ~]# rpm -qa freeipa-healthcheck
freeipa-healthcheck-0.14-1.fc38.noarch
[root@server ~]# ipa-healthcheck 
[]
[root@server ~]# pki-healthcheck 
[]
```
Traceback with 0.15:
```
[root@server ~]# rpm -qa freeipa-healthcheck
freeipa-healthcheck-0.15-1.fc38.noarch
[root@server ~]# ipa-healthcheck 
[]
[root@server ~]# pki-healthcheck 
Traceback (most recent call last):
  File "/usr/sbin/pki-healthcheck", line 33, in <module>
    sys.exit(load_entry_point('pkihealthcheck==0.1', 'console_scripts', 'pki-healthcheck')())
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/pki/server/healthcheck/core/main.py", line 53, in main
    sys.exit(checks.run_healthcheck())
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/site-packages/ipahealthcheck/core/core.py", line 468, in run_healthcheck
    api.Backend.ldap2.disconnect()
    ^^^^^^^^^^^
AttributeError: 'API' object has no attribute 'Backend'
```

The file `src/ipahealthcheck/core/core.py` was changed with commit 4a9ebb37cb296a41fefcd5898a3b1a69e6539720.

The issue was detected in the nightly test `testing-fedora/test_ipahealthcheck_cli_fsspace` in PR#[3134](https://github.com/freeipa-pr-ci2/freeipa/pull/3134), in `test_integration/test_ipahealthcheck.py::TestIpaHealthCLI::test_pki_healthcheck`.

PR #311, opened 2023-11-09
Remove call to api.Backend.ldap2.disconnect()

rcritten:
This was added while I was testing the IPA LDAP client cache performance. By disconnecting a summary of the cache is logged. I never intended it remain in the code.


flo-renaud:
@rcritten thanks for the patch, works for me. It properly fixes https://github.com/freeipa/freeipa-healthcheck/issues/310, you may want to add a reference to the issue in your commit msg.

rcritten:
Missing ticket reference added.

Issue #317, opened 2024-01-11
'DogtagCertsConfigCheck' fails, displaying the error message 'Malformed directive: ca.signing.certnickname=caSigningCert cert-pki-ca'


rcritten:
Cloned from downstream https://issues.redhat.com/browse/RHEL-21367

#### What were you trying to do that didn't work?

ipa-healthcheck displays CRITICAL error
Please provide the package NVR for which bug is seen:

    ipa-healthcheck-0.16-2.el9.noarch
    ipa-server-4.11.0-4.el9.x86_64
    idm-pki-ca-11.5.0-0.1.alpha4.el9.noarch

#### Steps to reproduce
```
 [root@master ~]# ipa-healthcheck
[
  {
    "source": "ipahealthcheck.dogtag.ca",
    "check": "DogtagCertsConfigCheck",
    "result": "CRITICAL",
    "uuid": "fa50649d-1af0-40b2-bc3b-94c71ab7924e",
    "when": "20240109102759Z",
    "duration": "0.241841",
    "kw": {
      "exception": "Malformed directive: ca.signing.certnickname=caSigningCert cert-pki-ca",
      "traceback": "Traceback (most recent call last):\n  File \"/usr/lib/python3.9/site-packages/ipahealthcheck/core/core.py\", line 56, in run_plugin\n    for result in plugin.check():\n  File \"/usr/lib/python3.9/site-packages/ipahealthcheck/core/plugin.py\", line 18, in wrapper\n    for result in f(*args, **kwds):\n  File \"/usr/lib/python3.9/site-packages/ipahealthcheck/dogtag/ca.py\", line 60, in check\n    val = get_directive(paths.CA_CS_CFG_PATH,\n  File \"/usr/lib/python3.9/site-packages/ipapython/directivesetter.py\", line 196, in get_directive\n    raise ValueError(\"Malformed directive: {}\".format(line))\nValueError: Malformed directive: ca.signing.certnickname=caSigningCert cert-pki-ca\n"
    }
  }
]
```
 
#### Expected results

The error in the check above should be fixed.

#### Actual results

Erroe: "Malformed directive: ca.signing.certnickname=caSigningCert cert-pki-ca"

This is related to PKI dropping the certificates from CS.cfg in 11.5.0. A parallel fix in IPA is being made to not fail if a value is not found:  https://pagure.io/freeipa/issue/9506

A related fix is needed in healthcheck to not report a false-positive:

[...]

PR #318, opened 2024-01-11
Skip DogtagCertsConfigCheck for PKI versions >= 11.5.0

rcritten:
In 11.5.0 the PKI project stopped storing the certificate blobs in CS.cfg. If we continue to check it we will report a false positive so skip it in that case.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/317


flo-renaud:
@rcritten thanks for the patch, works for me. Tested with dogtag-pki-server 11.3.1-1 and 11.5.0-0.1.alpha4 on fedora 38.

Issue #322, opened 2024-03-12
Improve reporting when checks time out


yrro:
```
# ipa-healthcheck --source=pki.server.healthcheck.clones.connectivity_and_data --check=ClonesConnectivyAndDataCheck  --output-type=human
    Unable to reach KRA at [https://ipa6.example.com:443](https://ipa6.example.com/): Request timed out
    ERROR: pki.server.healthcheck.clones.connectivity_and_data.ClonesConnectivyAndDataCheck
```

The 'request timed out' message implies that an attempt to fetch a URL on `ipa6` timed out. But that's not the case. The timeout is really caused by the 10 second limit on health checks. Let's improve the reporting in this case.

I've modified the plugin run code to produce something a bit clearer:

```
Command: pki -d /etc/pki/pki-tomcat/alias -C /tmp/tmp7nbqgra3/password.txt nss-cert-export --with-chain --format PEM Server-Cert cert-pki-ca /etc/pki/pki-tomcat/alias/ca.crt
Starting new HTTPS connection (1): ipa5.example.com:443
https://ipa5.example.com:443 "GET /kra/admin/kra/getStatus HTTP/1.1" 200 195
Content:
<?xml version="1.0" encoding="UTF-8" standalone="no"?><XMLResponse><State>1</State><Type>KRA</Type><Status>running</Status><Version>10.14.3-1.module+el8.8.0+18059+6d4394a9</Version></XMLResponse>
Status: running
KRA at https://ipa5.example.com:443 is running
Command: pki -d /etc/pki/pki-tomcat/alias -C /tmp/tmp4ej0vqt_/password.txt nss-cert-export --with-chain --format PEM Server-Cert cert-pki-ca /etc/pki/pki-tomcat/alias/ca.crt
--- Health check pki.server.healthcheck.clones.connectivity_and_data:ClonesConnectivyAndDataCheck cancelled after after 7 sec ---
  File "/bin/ipa-healthcheck", line 11, in <module>
    load_entry_point('ipahealthcheck==0.12', 'console_scripts', 'ipa-healthcheck')()
  File "/usr/lib/python3.6/site-packages/ipahealthcheck/core/main.py", line 52, in main
    sys.exit(ipachecks.run_healthcheck())
  File "/usr/lib/python3.6/site-packages/ipahealthcheck/core/core.py", line 475, in run_healthcheck
    config.timeout))
[...]

PR #323, opened 2024-03-12
Improved logging when plugins are cancelled for taking too long

yrro:
Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/322


rcritten:
Thanks for the patch. I'll try to review it soon.

rcritten:
I don't think including a traceback is necessary when a timeout occurs. I think this would lead to confusion in users thinking that this is a code issue and not merely a timeout. If it is persistent then users can debug logging to identify where it is timing out.

I set timeout=1 and ran --source ipahealthcheck.ipa.certs and the tracebacks don't really tell what is going on other than reads are failing (which isn't the root cause of the issue).

Issue #325, opened 2024-03-13
Need to change log file permissions as per CIS benchmark


tscherf:
As per CIS benchmark the log file permissions should be 640 but if we change /var/log/ipa-custodia.audit.log permissions to 640 then "ipa-healthcheck" reports a permission issue.

    ipa-healthcheck --failures-only
    [
      {
        "source": "ipahealthcheck.ipa.files",
        "check": "IPAFileCheck",
        "result": "ERROR",
        "uuid": "0abfd5fe-b7f7-4e3a-bac0-314f3a5f9c54",
        "when": "20240306081547Z",
        "duration": "0.002899",
        "kw": {       "key": "_var_log_ipa-custodia.audit.log_mode",       "path": "/var/log/ipa-custodia.audit.log",       "type": "mode",       "expected": "0644",       "got": "0640",       "msg": "Permissions of /var/log/ipa-custodia.audit.log are too restrictive: 0640 and should be 0644"     }

Below are the affected files:

/var/log/ipa-custodia.audit.log
/var/log/httpd/error_log 
/var/log/pki/pki-tomcat/ca/debug*

PR #326, opened 2024-03-13
Fixes log file permissions as per CIS benchmark

tscherf:
As per CIS benchmark the log file permissions should be 640 for some log files but if we change /var/log/ipa-custodia.audit.log permissions to 640 then "ipa-healthcheck" reports a permission issue.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/325


rcritten:
Code LGTM. Can you added signed-off-by to the commit? `git commit --amend -s`

Issue #340, opened 2024-10-08
IPAOpenSSLChainValidation changed behavior with ca-certificates update


flo-renaud:
The check IPAOpenSSLChainValidation from ipahealthcheck.ipa.certs changed behavior with the update of ca-certificates in rawhide.

Previously, if the root CA was missing from /etc/ipa/ca.crt, the check complained. It doesn't any more.

Scenario:
- install IPA with an externally-signed CA (2-step installation with --external-ca, sign CSR, provide cert with --external-cert-file)
- launch `ipa-healthcheck --source ipahealthcheck.ipa.certs --check IPAOpenSSLChainValidation`: no error detected
- remove the Root CA from /etc/ipa/ca.crt and launch again `ipa-healthcheck --source ipahealthcheck.ipa.certs --check IPAOpenSSLChainValidation`

With the previous version of ca-certificates, the command used to return the following error:
```
# rpm -qa ca-certificates
ca-certificates-2024.2.68_v8.0.302-2.fc41.noarch
# ipa-healthcheck --source ipahealthcheck.ipa.certs --check IPAOpenSSLChainValidation
[
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPAOpenSSLChainValidation",
    "result": "ERROR",
    "uuid": "262f9c42-e462-4b62-ade5-8b1bfd650953",
    "when": "20241008075949Z",
    "duration": "0.003525",
    "kw": {
      "key": "/var/lib/ipa/certs/httpd.crt",
      "reason": "O=IPA.TEST, CN=Certificate Authority\nerror 2 at 1 depth lookup: unable to get issuer certificate\nerror /var/lib/ipa/certs/httpd.crt: verification failed\n",
      "msg": "Certificate validation for {key} failed: {reason}"
    }
  },
  {
    "source": "ipahealthcheck.ipa.certs",
    "check": "IPAOpenSSLChainValidation",
    "result": "ERROR",
    "uuid": "1b3b328d-3da8-4ebc-9c25-6f60cbea8f86",
    "when": "20241008075949Z",
    "duration": "0.006803",
    "kw": {
      "key": "/var/lib/ipa/ra-agent.pem",
      "reason": "O=IPA.TEST, CN=Certificate Authority\nerror 2 at 1 depth lookup: unable to get issuer certificate\nerror /var/lib/ipa/ra-agent.pem: verification failed\n",
      "msg": "Certificate validation for {key} failed: {reason}"
    }
[...]

PR #341, opened 2024-10-09
IPAOpenSSLChainValidation: ignore default trust store

flo-renaud:
The check IPAOpenSSLChainValidation is ensuring that the whole certification chain is present in IPA for httpd and RA certificates.
It internally calls openssl verify -CAfile /etc/ipa/ca.crt.

With the latest version of ca-certificates package, openssl verify also uses the default trust store. Since the test wants to check the chain presence in /etc/ipa/ca.crt, add the -no-CAfile -no-CApath and -no-CAstore options to ensure that only /etc/ipa/ca.crt is used as trusted source.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/340

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -1074,6 +1074,7 @@ class IPAOpenSSLChainValidation(IPAPlugin):
                 '-verbose',
                 '-show_chain',
                 '-CAfile', paths.IPA_CA_CRT,
+                '-no-CAfile', '-no-CApath', '-no-CAstore',
                 file]
 
         return ipautil.run(args, raiseonerr=False)


rcritten:
I believe that this is due to ca-certificates adding directory hash links to /etc/pki/tls/certs and dropping the PEM bundle file. This was introduced in ca-certificates-2024.2.69_v8.0.303-5. The bundle will be restored (it is in distgit now but isn't built/released yet).

This code worked previously because a specific CAfile was specified so that the global trust wasn't loaded, only the provided file. With these hash directories in place the global trust is loaded via those.

So I guess my ask is that for historical purposes you mention at least that this affects Fedora 42/rawhide.

flo-renaud:
Updated the commit message:
```
IPAOpenSSLChainValidation: ignore default trust store

The check IPAOpenSSLChainValidation is ensuring that the
whole certification chain is present in IPA for httpd and RA
certificates.
It internally calls openssl verify -CAfile /etc/ipa/ca.crt.

With the latest version of ca-certificates package shipped in
rawhide/Fedora 42, openssl verify also uses the default trust
store. Since the test wants to check the chain presence in
/etc/ipa/ca.crt, add the -no-CAfile -no-CApath and -no-CAstore
options to ensure that only /etc/ipa/ca.crt is used as trusted
source.
```

rcritten:
ack, thanks.

PR #349, opened 2025-02-17
Replace fips-mode-setup

duzda:
RHEL10 doesn't support fips-mode-setup, therefore this call has been replaced with simple reading of a proc file. The tests have been edited accordingly, inconsistent has been removed and instead replaced by a test supplying an arbitrary value, that should never occur.

Fixes: #350

### DIFF
--- a/src/ipahealthcheck/meta/core.py
+++ b/src/ipahealthcheck/meta/core.py
@@ -5,6 +5,7 @@
 import logging
 import os
 import socket
+from pathlib import Path
 from ipahealthcheck.core import constants
 from ipahealthcheck.core.exceptions import TimeoutError
 from ipahealthcheck.core.plugin import Result, duration
@@ -13,8 +14,8 @@ from ipapython import ipautil
 from ipapython.version import VERSION, API_VERSION
 from ipaplatform.paths import paths
 
-if 'FIPS_MODE_SETUP' not in dir(paths):
-    paths.FIPS_MODE_SETUP = '/usr/bin/fips-mode-setup'
+if 'PROC_FIPS_ENABLED' not in dir(paths):
+    paths.PROC_FIPS_ENABLED = '/proc/sys/crypto/fips_enabled'
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/meta/core.py
```diff
@@ -25,31 +26,30 @@ class MetaCheck(Plugin):
...
 
         rval = constants.SUCCESS
-        if not os.path.exists(paths.FIPS_MODE_SETUP):
-            fips = "missing {}".format(paths.FIPS_MODE_SETUP)
-            logger.debug('%s is not installed, skipping',
-                         paths.FIPS_MODE_SETUP)
+        if not os.path.exists(paths.PROC_FIPS_ENABLED):
+            fips = "missing {}".format(paths.PROC_FIPS_ENABLED)
+            logger.debug("Can't find %s, skipping" %
+                         paths.PROC_FIPS_ENABLED)
```
rcritten, line 32: I'm a bit torn on this related to other distributions, whether enforcing that the file exist would break them.
When I originally wrote this I didn't want to hold it against a user that they didn't have a tool installed. But the file in /proc should be created by the kernel so if that's missing it points to a larger issue.
I think we should set rval to WARNING in this case. This file is created by the kernel so if FIPS is disabled in the KERNEL that seems odd. A user can suppress the warning if indeed they have this use-case.

src/ipahealthcheck/meta/core.py
```diff
@@ -25,31 +26,30 @@ class MetaCheck(Plugin):
...
         else:
             try:
-                result = ipautil.run([paths.FIPS_MODE_SETUP,
-                                      '--is-enabled'],
-                                     capture_output=True,
-                                     raiseonerr=False,)
+                proc_fips_enable_path = Path(paths.PROC_FIPS_ENABLED)
+                result_text = proc_fips_enable_path.read_text()
+                result = int(result_text)
```
rcritten, line 37: There is a platforms call in freeipa that would do much of this for you, is_fips_enabled(). Now in that we consider FIPS not enabled if the file doesn't exist so we could alternatively take that stance as well. It would be consistent anyway.
I added the timeout part in case executing the command took a while. It wasn't exactly fast. I think that part can be dropped. And the freeipa version of this just looks for value != 0. So I think this can be simplified by using the same call as freeipa itself does.

rcritten: Then again, the freeipa test is a hammer while the healthcheck test is more nuanced. I guess I could go either way.

src/ipahealthcheck/meta/core.py
duzda: I think we can make the assumption, that this file exists, otherwise, the user is using a custom kernel and is probably not thinking about FIPS at all. Nevertheless, the return value is a warning now.

src/ipahealthcheck/meta/core.py
duzda: You're right, there's the `is_fips_enabled` in the `BaseTaskNamespace` class, it would be probably cleaner that way.

duzda: Ultimately I've decided to keep it this way, just got rid of the timeout. Even though most of the errors are very unlikely, we should provide more user feedback than just *enabled*/*disabled* if something goes really wrong.

src/ipahealthcheck/meta/core.py
```diff
@@ -13,8 +14,8 @@
 from ipapython.version import VERSION, API_VERSION
 from ipaplatform.paths import paths
 
-if 'FIPS_MODE_SETUP' not in dir(paths):
-    paths.FIPS_MODE_SETUP = '/usr/bin/fips-mode-setup'
+if 'PROC_FIPS_ENABLED' not in dir(paths):
+    paths.PROC_FIPS_ENABLED = '/proc/sys/crypto/fips_enabled'
```
rcritten, line 18: PROC_FIPS_ENABLED is in ipaplatform/base.py so should be visible in any derived platform. One can override a path but not remove it so I don't think this part is necessary. It's good defensive coding though.

tests/test_meta.py
```diff
@@ -10,11 +10,10 @@
 from ipahealthcheck.core import config, constants
 from ipahealthcheck.meta.plugin import registry
 from ipahealthcheck.meta.core import MetaCheck
-from ipapython import ipautil
 from ipaplatform.paths import paths
 
-if 'FIPS_MODE_SETUP' not in dir(paths):
-    paths.FIPS_MODE_SETUP = '/usr/bin/fips-mode-setup'
+if 'PROC_FIPS_ENABLED' not in dir(paths):
+    paths.PROC_FIPS_ENABLED = '/proc/sys/crypto/fips_enabled'
```
rcritten, line 16: similar platforms comment.

src/ipahealthcheck/meta/core.py
duzda: Thanks for letting me know!

Issue #353, opened 2025-03-24
Notify user about wrong umask


duzda:
We should give proper warning message to the user when wrong umask is set. The current solution might be misleading and doesn't really hint towards the issue.

```
ERROR: ipahealthcheck.ipa.files.IPAFileCheck._run_ipa_services.list_mode: Permissions of /run/ipa/services.list are too restrictive: 0640 and should be 0644 
```

Instead, we should tell user to change their umask.

PR #354, opened 2025-03-25
Warn about unexpected umask

duzda:
We should let users know, that their umask set wrong to an unsupported value, it does not make much sense to check file permissions when wrong umask is set.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/353

### DIFF
--- a/src/ipahealthcheck/core/files.py
+++ b/src/ipahealthcheck/core/files.py
@@ -12,6 +12,19 @@ from ipahealthcheck.core.plugin import Result, duration
 
 logger = logging.getLogger()
 
+EXPECTED_UMASK = 0o022
+
+
+def current_umask():
+    """
+    Retrieves current umask by setting it temporarily
+
+    :returns: int umask
+    """
+    umask = os.umask(EXPECTED_UMASK)
+    os.umask(umask)
+    return umask
+
 
 class FileCheck:
     """Generic check to validate permission and ownership of files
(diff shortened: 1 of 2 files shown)


src/ipahealthcheck/core/files.py
```diff
@@ -12,6 +12,19 @@
...
+
+
+def current_umask():
+    """
+    Retrieves current umask by setting it temporarily
+
+    :returns: int umask
+    """
+    umask = os.umask(0)
+    os.umask(umask)
```
rcritten, line 25: I'd  prefer if you'd set it to 022 instead of 0 because if something were to go wrong the permissions would be wide open. It is probably perfectly safe and healthcheck is single-threaded it just gives me the heebie jeebies.

tests/test_core_files.py
```diff
@@ -303,3 +303,20 @@ def test_ipa_files_format(mock_pkinit):
 
     for result in results.results:
         assert result.result in (constants.SUCCESS, constants.WARNING)
+
+
+@patch('os.umask')
+def test_umask(mock_umask):
+    mock_umask.return_value = 0o027
```
rcritten, line 310: Not to be too pedantic but can you change the test name to test_bad_umask? That's what this does.

duzda: Not pedantic at all, I just forgot to check

Issue #362, opened 2025-07-11
Issue an ERROR if 389 or dirsrv is configured in SSSD -> pac -> allowed_uids


rcritten:
The issue here is that during a SASL/GSSAPI bind 389-ds send the PAC from the Kerberos ticket of the machine account trying to connect. The PAC responder tried to evaluate the SIDs in PAC by connecting to the SSSD backend which might have to read the data from 389-ds (or even has to do a SASL/GSSAPI bind on its own first) which cause additional load on 389-ds and in a heavily load situation 389-ds was more or less stuck.

Since the searches are currently failing because SIDs which are looked up are for "Domain Controllers" and "Domain Computer" and those are not defined in IPA SSSD will try to look them up constantly because the default negative cache timeout is only 10s.

So error out if 389/dirsrv (DS_USER) are in allowed_uids.

PR #363, opened 2025-07-14
Check that allowed_uids in the SSSD config is valid

rcritten:
If UID 389 or user dirsrv is found in 'allowed_uids' in sssd.conf section [pac] then this will prevent SSSD from resolving the LDAP service account locally and can cause IPA services to fail.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/362


tests/test_ipa_config.py
```diff
@@ -52,3 +88,108 @@ def test_last_success_enabled(self):
...
+        self.results = capture_results(f)
+
+        assert len(self.results) == 1
+        result = self.results.results[0]
+        assert result.result == constants.SUCCESS
+        assert result.source == 'ipahealthcheck.ipa.config'
+        assert result.check == 'SSSDAllowedUids389Check'
+
+    @patch('SSSDConfig.SSSDConfig')
+    def test_sssd_ok_multiple_allowed_uids_configured(self, mock_sssd):
```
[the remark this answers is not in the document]
rcritten: This is what I get for adding a test at the last second before pushing.

src/ipahealthcheck/ipa/config.py
```diff
@@ -44,3 +54,56 @@ def check(self):
...
+            uids = service.get_option('allowed_uids')
+        except SSSDConfig.NoOptionError:
+            logger.debug('ok, allowed_uids is undefined')
+            yield Result(self, constants.SUCCESS, key='SSSD_allowed_uids')
+            return
+        else:
+            uids = {s.strip() for s in uids.split(',') if s.strip()}
+            candidates = {'389', DS_USER}
+            invalid = uids.intersection(candidates)
+            if invalid:
```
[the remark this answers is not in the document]
frasertweedale: I didn't know python grew this feature.  Is nothing sacred? :sweat_smile:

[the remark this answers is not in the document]
abbra: It is Python 3.8+, so probably not yet because RHEL 8 is python 3.6. https://peps.python.org/pep-0572/

src/ipahealthcheck/ipa/config.py
```diff
@@ -44,3 +51,56 @@ def check(self):
...
+                yield Result(
+                    self, constants.ERROR,
+                    key='SSSD_allowed_uids',
+                    invalid=', '.join(invalid),
+                    msg="User/UID invalid} found in 'allowed_uids' in "
```
frasertweedale, line 101: > Output raised ValueError: Single '}' encountered in format string

(missing opening brace around `invalid`)

src/ipahealthcheck/ipa/config.py
```diff
@@ -44,3 +51,56 @@ def check(self):
                 constants.SUCCESS,
                 key='krbLastSuccessfulAuth'
             )
+
+
+class SSSDAllowedUids389Check(IPAPlugin):
```
frasertweedale, line 56: missing `@registry` decorator:
```python
@registry
class SSSDAllowedUids389Check(IPAPlugin):
```

src/ipahealthcheck/ipa/config.py
```diff
@@ -44,3 +51,56 @@ def check(self):
...
+
+        try:
+            service = sssdconfig.get_service('pac')
+        except SSSDConfig.NoServiceError:
+            yield Result(
+                    self, constants.ERROR,
+                    key='SSSD_allowed_uids',
+                    msg="No [pac] section defined in sssd.conf"
+            )
+            return
```
frasertweedale, line 84: I think we should allow this case and treat it the same as an empty `[pac]` section.  SSSD runs fine when the section is not defined.

frasertweedale (changes requested):

src/ipahealthcheck/ipa/config.py
rcritten: Fixed

frasertweedale (approved):
Ack.  For the no `[pac]` section case, I'll leave the final decision with you.

Issue #342, opened 2024-10-17
running ipa-healthcheck --failures-only  on newly installed ipa-server lists ERROR for IPACertfileExpirationCheck


4gemenot:
I'm getting a list of unable to find certificate.

ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072907: Request id 20240828072907: Unable to retrieve cert 'auditSigningCert cert-pki-ca' from '/etc/pki/pki-tomcat/alias': Unable to find certificate
ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072908: Request id 20240828072908: Unable to retrieve cert 'ocspSigningCert cert-pki-ca' from '/etc/pki/pki-tomcat/alias': Unable to find certificate
ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072909: Request id 20240828072909: Unable to retrieve cert 'subsystemCert cert-pki-ca' from '/etc/pki/pki-tomcat/alias': Unable to find certificate
ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072910: Request id 20240828072910: Unable to retrieve cert 'caSigningCert cert-pki-ca' from '/etc/pki/pki-tomcat/alias': Unable to find certificate
ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072911: Request id 20240828072911: Unable to retrieve cert 'Server-Cert cert-pki-ca' from '/etc/pki/pki-tomcat/alias': Unable to find certificate
ERROR: ipahealthcheck.ipa.certs.IPACertfileExpirationCheck.20240828072913: Request id 20240828072913: Unable to retrieve cert 'Server-Cert' from '/etc/dirsrv/slapd-IDM-SEMAT-GOV-SA': Unable to find certificate

Running 'getcert list' all certificates show with status MONITORING,

flo-renaud:
Can you provide the output of `ipa-healthcheck  --source ipahealthcheck.ipa.certs --check IPACertfileExpirationCheck --verbose --debug` ?

4gemenot:
[root@s1biok20idmp01 ~]# ipa-healthcheck --source ipahealthcheck.ipa.certs --check IPACertfileExpirationCheck --verbose --debug 
Loading StateFile from '/var/lib/ipa/sysrestore/sysrestore.state'
[...]

PR #364, opened 2025-07-16
Check that the expected NSS token matches the current FIPS state

rcritten:
In order to be FIPS-compliant, and actually be able to renew certificates, the NSS token name needs to match the expectation based on the current FIPS state. The token name will change depending on the state.

So if a user installs IPA and then switches to FIPS mode:
1. The system is non-compliant
2. Renewals will fail because the NSS Certificate DB token is not present in the database.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/342

## Summary by Sourcery

Add a new healthcheck to verify certmonger NSS token names against the current FIPS state and update tests to cover various token scenarios and HSM skips.

New Features:
- Introduce CertmongerFIPTokensCheck plugin to validate that NSS certmonger token names match the expected value for the current FIPS mode

Bug Fixes:
- Correct the result indexing in an existing test to use the last result instead of the first

Tests:
- Add tests for non-FIPS and FIPS token correctness, wrong token detection, and HSM token skip behavior

### DIFF
--- a/src/ipahealthcheck/ipa/certs.py
+++ b/src/ipahealthcheck/ipa/certs.py
@@ -22,6 +22,7 @@ from ipalib import x509
 from ipalib.install import certmonger
 from ipalib.constants import RENEWAL_CA_NAME, IPA_CA_RECORD
 from ipaplatform.paths import paths
+from ipaplatform.tasks import tasks
 from ipaserver.install import certs
 from ipaserver.install import dsinstance
 from ipaserver.install import krainstance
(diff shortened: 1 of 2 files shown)


tests/test_ipa_tracking.py
```diff
@@ -128,3 +132,165 @@ def test_one_stuck(self):
...
+    def test_nonfips_token_correct(self, mock_value, mock_fips):
+        mock_value.side_effect = [
+            'FILE',
+            'FILE',
+            'NSSDB', 'NSS Certificate DB',
+        ]
+        mock_fips.return_value = False
+        set_requests(add=self.nss_tracking)
+#        mock_requests.return_value=get_expected_requests()
+#        mock_obj.side_effect=create_mock_dbus
```
frasertweedale, line 173: Can we tidy this up, or leave a comment to explain why these lines are commented out?

src/ipahealthcheck/ipa/certs.py
```diff
@@ -1568,3 +1569,68 @@ def check(self):
 
         if len(requests) == 0:
             yield Result(self, constants.SUCCESS, key='no_stuck')
+
+
+@registry
+class CertmongerFIPTokensCheck(IPAPlugin):
```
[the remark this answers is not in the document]
rcritten: Yeah, yuck. Fixed.

rcritten:
I incorporated your test suggestion. Thanks for doing the leg work to figure that out.

tests/test_ipa_tracking.py
frasertweedale: Still got this (apparently?) vestigal commented-out code.  Not a blocker.  Leave it to you whether to remove or not.  (Maybe it's there for a reason...)

rcritten:
Commented lines removed.

PR #370, opened 2025-09-08
add ipa-certupdate hint for missing LWCA tracking request

frasertweedale:
Missing LWCA tracking requests is an expected scenario.  Admins need to run `ipa-certupdate` to set up the tracking requests.  Include this hint in the error message.

Drive-by change, no ticket.


rcritten:
I think I'd prefer it if we try to determine if the given certificate is a LWCA rather than always appending this. It could cause unnecessary running of ipa-certupdate to not fix the problem. I suspect a regex on the nickname would handle this pretty well.

I also uncovered a related issue but in IPA. Deleting a LWCA and re-running ipa-certupdate does not remove the tracking which will also spur healthcheck errors.

PR #379, opened 2026-05-04
Correct the logic on trust checking in IPATrustPackageCheck

### LINKED ISSUE #378: IPATrustPackageCheck skipped message is incorrect
IPATrustPackageCheck displays msg 'Skipped. Not a trust controller' on IPA Server with trust added. Although the IPA server has trust established and all the other checks are displaying correct message the IPATrustPackageCheck displays incorrect message.

```{
    "source": "ipahealthcheck.ipa.trust",
    "check": "IPATrustPackageCheck",
    "result": "SUCCESS",
    "uuid": "998bd14a-a25b-487b-ba77-8ef19b9d6b79",
    "when": "20260113110456Z",
    "duration": "0.000018",
    "kw": {
      "msg": "Skipped. Not a trust controller"
    }```

Downstream issue https://redhat.atlassian.net/browse/RHEL-140929

rcritten:
The check would be skipped if either not a trust controller or
a trust agent. It should skip it if both conditions are true.

This resulted in an incorrect message in the output because it
skipped if the server was a controller. If it was not then
the function exited without checking to see if it was an agent
instead.

Fixes: https://github.com/freeipa/freeipa-healthcheck/issues/378

Also update the Fedora versions to test against.

## Summary by Sourcery

Fix trust role detection in IPATrustPackageCheck and update CI Fedora test matrix.

Bug Fixes:
- Correct trust controller/agent check to only skip when the server is neither a trust controller nor a trust agent, preventing incorrect skip messages.

CI:
- Update GitHub Actions Fedora release matrix to test against Fedora 43 and 44.

### DIFF
--- a/.github/workflows/pipelines.yml
+++ b/.github/workflows/pipelines.yml
@@ -35,7 +35,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        fedora-release: [41, 42]
+        fedora-release: [43, 44]
 
     steps:
     - uses: actions/checkout@v4
--- a/src/ipahealthcheck/ipa/trust.py
+++ b/src/ipahealthcheck/ipa/trust.py
@@ -685,15 +685,13 @@ class IPATrustPackageCheck(IPAPlugin):
     """
     @duration
     def check(self):
-        if self.registry.trust_controller:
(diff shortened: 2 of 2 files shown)


rjeffman (approved):
LGTM. Fixes the referenced issue.

